PR #2, opened 2020-03-13
Send email in async mode and fix redirect issue with ssl

akondas:


ansible/roles/app.env/templates/.env.local
```diff
@@ -1,4 +1,5 @@
 APP_HOST={{ server_hostname }}
+APP_PUBLIC_HOST={{ lookup('env', 'REPMAN_APP_PUBLIC_HOST') or server_hostname }}
```
akondas: this is only for our public instance, which will be hosted on `app.` subdomain

PR #3, opened 2020-03-13
Handle invalid package name and source path; Remove dists directory

karniv00l:


tests/Unit/Service/PackageSynchronizer/ComposerPackageSynchronizerTest.php
```diff
@@ -84,12 +86,26 @@ public function testSynchronizePackageWithToken(): void
...
         $json = unserialize((string) file_get_contents($path));
         self::assertTrue($json['packages']['repman-io/repman'] !== []);
         @unlink($path);
     }
+
+    public function testSynchronizePackageWithInvalidName(): void
+    {
+        $this->synchronizer->synchronize(PackageMother::withOrganization('path', $this->resourcesDir.'path/invalid-name', 'buddy'));
+        // exception was not thrown
+        self::assertTrue(true);
```
akondas: we need to check if correct error was handled

src/Service/Organization/PackageManager.php
```diff
@@ -8,16 +8,19 @@
...
 {
     private Storage $distStorage;
     private string $baseDir;
+    private Filesystem $filesystem;
 
     public function __construct(Storage $distStorage, string $baseDir)
     {
         $this->distStorage = $distStorage;
         $this->baseDir = $baseDir;
+        $this->filesystem = new Filesystem();
```
akondas: why not dependency injection?

src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
```diff
@@ -54,7 +56,16 @@ public function synchronize(Package $package): void
                 }
             }
 
+            if (!$latest) {
+                throw new \RuntimeException('Package not found');
+            }
```
akondas: this can be check earlier, right after `$packages = $repository->getPackages()`, for example:
```php
if($packages === []) {
}
```

src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
```diff
@@ -54,7 +56,16 @@ public function synchronize(Package $package): void
...
+
             $name = $latest->getPrettyName();
+
+            if (!preg_match(self::NAME_PATTERN, $name, $matches) || empty($matches)) {
```
akondas: i think we can move this also to `Package` and protect domain from setting incorrect package name

PR #10, opened 2020-03-18
Fix warning on user profile page when emial is not verified

akondas:


templates/user/profile.html.twig
```diff
@@ -7,7 +7,7 @@
 {% if not is_user_confirmed %}
     <h3>Verify your email address</h3>
     <div class="alert alert-warning">
-        Please click the activation link for <strong>{{ "asd@asd.pl" }}</strong> to verify your email.
```
akondas: :scream:

PR #17, opened 2020-03-25
Replace file writes with atomic file writer for better conccurency support

akondas:


src/Service/AtomicFile.php
```diff
@@ -0,0 +1,16 @@
...
+namespace Buddy\Repman\Service;
+
+final class AtomicFile
+{
+    public static function write(string $filename, string $content): void
+    {
+        $tempFilename = $filename.'.'.uniqid('temp', true);
+
+        file_put_contents($tempFilename, $content);
+        rename($tempFilename, $filename);
```
akondas: reference: https://tqdev.com/2018-locking-file-cache-php

PR #35, opened 2020-04-07
Require user current password on password change

akondas:


templates/user/profile.html.twig
```diff
@@ -6,41 +6,49 @@
...
+        <button
+            class="btn btn-danger"
+            type="button"
+            data-target="confirmation"
+            data-action="{{ url('user_remove') }}"
+            data-method="DELETE"
+        >
+            Delete Account
+        </button>
+    </div>
```
akondas, line 53: Screenshot from 2020-04-07 08-57-01

PR #49, opened 2020-04-14
Allow admin to update user roles (promote to admin)

akondas:


src/Query/Admin/UserQuery/DbalUserQuery.php
```diff
@@ -22,7 +22,7 @@ public function findAll(int $limit = 20, int $offset = 0): array
     {
         return array_map(function (array $data): User {
             return $this->hydrateUser($data);
-        }, $this->connection->fetchAll('SELECT id, email, status, roles FROM "user" LIMIT :limit OFFSET :offset', [
+        }, $this->connection->fetchAll('SELECT id, email, status, roles FROM "user" ORDER BY email LIMIT :limit OFFSET :offset', [
```
akondas, line 25: prevent user list to flicker :wink:

src/Entity/User.php
```diff
@@ -189,6 +189,14 @@ public function getRoles(): array
         return array_unique($roles);
     }
 
+    /**
+     * @param string[] $roles
+     */
+    public function changeRoles(array $roles): void
+    {
+        $this->roles = array_unique($roles);
```
marmichalski, line 197: Should add array_values here to prevent the serialized value to transform into json object 🤷

PR #50, opened 2020-04-15
Stop propagation of expired token exception

akondas:
this prevents the caught exception from being sent to the sentry


src/Service/Symfony/RefreshOAuthTokenListener.php
```diff
@@ -38,15 +38,16 @@ public function onKernelException(ExceptionEvent $event): void
         }
 
         $event->setResponse(new RedirectResponse($this->router->generate('refresh_oauth_token', ['type' => $exception->type()])));
+        $event->stopPropagation();
```
marmichalski, line 41: `setResponse` already stops propagation :scream:

akondas: :thinking: actually yes, this was added with momentum, but the prudent always insured :surfer:

Issue #56, opened 2020-04-16
Organization members


akondas:
Currently, the organization has only the owner - the user who created it. 

Members - will give the opportunity to manage people in the organization, send invitations and remove them.

Features:
- [ ] invite by e-mail
- [ ] two manage scope: member, admin

akondas:
Then let's try to write the permissions:

**Member**
- manage packages (can't remove)
- see organization tokens

**Admin**
- can remove package
- manage tokens
- organization configuration 

**Owner**
- remove organization
- transfer ownership

akondas:
Now is time form flow:

- Owner creates invitation
  - choose a role
  - invitation has a finite period of validity
- User accepts invitation
  - already exist: click link in e-mail 
  - new users: creates account

karniv00l:
looks good

akondas:
:tada: released in [0.2.0](https://github.com/repman-io/repman/releases/tag/0.2.0)

PR #57, opened 2020-04-16
Allow invite user to organization (application layer)

### LINKED ISSUE #53: Should add array_values here to prevent the serialized value to transform into json object 🤷
Should add array_values here to prevent the serialized value to transform into json object 🤷

_Originally posted by @marmichalski in https://github.com/repman-io/repman/pull/49_

akondas:
Resolves #53

### DIFF
--- a/src/Entity/Organization.php
+++ b/src/Entity/Organization.php
@@ -4,11 +4,14 @@ declare(strict_types=1);
 
 namespace Buddy\Repman\Entity;
 
+use Buddy\Repman\Entity\Organization\Invitation;
+use Buddy\Repman\Entity\Organization\Member;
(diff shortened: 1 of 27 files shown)


src/Entity/Organization.php
```diff
@@ -157,4 +174,28 @@ public function changeAlias(string $alias): void
     {
         $this->alias = $alias;
     }
+
+    public function inviteUser(string $email, string $role, string $token): void
+    {
+        if ($this->invitations->exists(fn (int $key, Invitation $invitation) => $invitation->email() === $email)) {
+            return;
```
akondas, line 181: instead of raising an exception, I'd like to try a different approach

akondas:
Part of #56 (application layer)

PR #60, opened 2020-04-17
Add command for syncing packagist releases

karniv00l:


src/Command/ProxySyncReleasesCommand.php
```diff
@@ -0,0 +1,77 @@
...
+            list($name, $version) = explode(' ', (string) $item->guid);
+            $this->syncPackages($proxy, $name, $version);
+        }
+
+        return 0;
+    }
+
+    private function syncPackages(Proxy $proxy, string $name, string $version): void
+    {
+        foreach ($proxy->syncedPackages() as $package) {
```
akondas, line 54: what about first dump synced packages to array with name as keys, and then use only `isset` to prevent looping again for each item

docker-compose.yml
```diff
@@ -39,6 +39,17 @@ services:
     depends_on:
       - app
 
+  cron:
```
akondas, line 42: we also need to create a role for Ansible

Issue #59, opened 2020-04-17
Support for self-hosted GitLab


mgrajcarek:
Hi!

I have another feature request. 
Could you add a possibility to setup GitLab URL, so I could use private, self hosted GitLab instance istead of gitlab.com? 
It could be configured by .env variable, nothing funcy needed.
Thanks in advance!

Best, Maciej

akondas:
Currently, this can be done via SSH, you choose the Git package type, you give the url of your repository and you configure your instance to let the repman in. 

In this case, however, you will need to configure the webhook manually, so the option you mention will be useful and will be done.

mgrajcarek:
SSH configuration works as a charm!
Looking forward for extension of this feature in a near future.

JonEllis:
I have the bodgiest "fix" for this - you can simply replace the usages of `gitlab.com` in php files in the installed app (not just this repo). I should say this is not a good fix, but a quick and dirty one.

Assuming you've set your `OAUTH_GITLAB_CLIENT_ID` and `OAUTH_GITLAB_CLIENT_SECRET` env vars, you can run this to replace the Gitlab instance (replace `gitlab.example.com` with your Gitlab instance):

``` bash
find . -type f -name '*.php' -print0 | xargs -0 grep -l gitlab.com | xargs sed -i 's/gitlab.com/gitlab.example.com/g'
```

If you're deploying using docker (like I am), you can add override the command of the app container by adding this to your `docker-compose-override.yml` file to replace it before the app starts:

``` yaml
version: '3.4'

services:
  app:
    command: >
      bash -c "
      find . -type f -name '*.php' -print0 | xargs -0 grep -l gitlab.com | xargs sed -i 's/gitlab.com/gitlab.example.com/g' &&
      bin/console cache:clear &&
      bin/console cache:warmup &&
      bin/console d:m:m --no-interaction &&
      bin/console messenger:setup-transports --no-interaction &&
      php-fpm"
```

Now, this isn't perfect, but lets you authenticate using your Gitlab instance and find repositories through the nice integration.

A caveat is that it generates HTTPS rather than SSH repos, so you have to "fix" them in the database.

Then you can set the `GIT_SSH_COMMAND=ssh -i /app/ssh/repman.key` env variable and repman will use the key when performing SSH git operations.


tl;dr - It's probably best to wait until this project is updated to support self-hosted instances and to handle SSH keys (there's another issue for that bit already).

akondas:
[...]

PR #63, opened 2020-04-20
Extract GitLab API url to allow to integrate with self-hosted instance

akondas:


akondas:
Fixes #59 (although we will probably do it even better)

PR #66, opened 2020-04-20
Run docker pipelines after tests, lock versions

karniv00l:


docker-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
 
   consumer:
     restart: always
-    image: buddy/repman
+    image: buddy/repman:0.1.0
```
akondas, line 34: this should be done before tag

PR #70, opened 2020-04-21
Use lock to prevent multiple jobs run simultaneously

karniv00l:


src/Command/ProxySyncReleasesCommand.php
```diff
@@ -39,6 +43,11 @@ protected function configure()
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        $lock = $this->lockFactory->createLock('packagist_releases_feed');
+        if (!$lock->acquire()) {
+            return 0;
+        }
+
```
karniv00l, line 50: lock is released automatically on instance destruction

akondas:
we need to add settings for `codecov/patch` like for `codecov/project`

akondas:
```
coverage:
    status:
        project:
            default:
                informational: true
        path:
            default:
                informational: true
```

src/Command/ProxySyncReleasesCommand.php
```diff
@@ -59,6 +77,7 @@ private function syncPackages(\SimpleXMLElement $feed): void
         }
 
         foreach ($feed->channel->item as $item) {
+            $this->lock->refresh();
```
akondas, line 80: I think `refresh` can happen in `if (isset($syncedPackages[$name])) {` condition

karniv00l: done

PR #77, opened 2020-04-22
Mailer setup - new section in Readme

mgrajcarek:
Added Configuration section and first element of it - Mailer setup


akondas:
Thanks @mgrajcarek

PR #79, opened 2020-04-22
Lowercase email during User initialization

mgrajcarek:
Since emails are case insensitive, they should be lowercased.
Currently multiple accounts can be created using one, single email.

### DIFF
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -93,7 +93,7 @@ class User implements UserInterface
     public function __construct(UuidInterface $id, string $email, string $emailConfirmToken, array $roles)
     {
         $this->id = $id;
-        $this->email = $email;
+        $this->email = \mb_strtolower($email);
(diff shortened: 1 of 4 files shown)


src/Entity/User.php
```diff
@@ -93,7 +93,7 @@ class User implements UserInterface
     public function __construct(UuidInterface $id, string $email, string $emailConfirmToken, array $roles)
     {
         $this->id = $id;
-        $this->email = $email;
+        $this->email = \mb_strtolower($email);
```
akondas, line 96: this should also be added to unique email validator `Buddy\Repman\Validator\UniqueEmailValidator` to handle this case in UI layer

mgrajcarek: OK!

mgrajcarek: @akondas done

PR #86, opened 2020-04-22
Use lowercased email during registration and authentication

karniv00l:


src/Security/UserGuardHelper.php
```diff
@@ -29,7 +29,7 @@ public function userExists(string $email): bool
     public function authenticateUser(string $email, Request $request): void
     {
         $this->guardHandler->authenticateWithToken(
-            $this->authenticator->createAuthenticatedToken($this->users->getByEmail($email), 'main'),
+            $this->authenticator->createAuthenticatedToken($this->users->getByEmail(\strtolower($email)), 'main'),
```
karniv00l, line 32: sentry issue `REPMAN-V`

akondas, line 32: `mb_strtolower`

src/MessageHandler/User/CreateUserHandler.php
```diff
@@ -26,7 +26,7 @@ public function __invoke(CreateUser $message): void
     {
         $user = new User(
             Uuid::fromString($message->id()),
-            $message->email(),
+            \strtolower($message->email()),
```
akondas, line 29: it can be moved to `User` constructor

PR #87, opened 2020-04-22
WIP: Mysql Support

sadortun:


akondas:
Can we know the motivation for these changes? I'm not opposed, but it will complicate the number of tests considerably, because you will have to test two different stacks. Since PostgreSQL is provided by both docker image and ansible role, why support for two different databases?

sadortun:
Hi @akondas 

In general, i  think that MySQL is widely used by a lot of organizations and it would be sad not to support it.

In more specifics, we run most of of our workloads on MySQL, and our infrastructure to manage migrations, updates and backups is all setup on MySQL. And i think a lot of organizations also.

From what i am currently testing there is only very minor changes to be done to support both.  

The best would be of course to use DQL to build the queries to ensure they are 100% compatible. But from what i see, you dont use complex SQL or PostgreSQL specific functions ( except a very few)

akondas:
I understand that in this case your approach only applies to manual application deployment? From what I've seen, however, most use docker images (although this is just the beginning of the project).

sadortun:
Not really.  We only use automated deployments, but still, we need to manage migrations between database versions, data persistence, and backup managements.  You are less at risk of problem if you can use a single type instead of two.

akondas:
Hey @sadortun, are you going to continue working on it or maybe we will close this PR? It seems to me that at the moment docker-compose or ansible allows you to effectively manage the infrastructure and there is no need to implement another database engine. Let me know what you think about it :wink:

sadortun:
Hi @akondas 

Unless there is more community support for MySql, i wont support this by myself.

So, yes, we can close, and hope that someone else will be interested :) 

Have a good day,
Samuel

akondas:
Thanks for quick response.

sadortun:
hehe, i am always online! I sleep one hour every 9 days 😆

PR #95, opened 2020-04-23
Add "origin" param when redirecting after oauth account registration

karniv00l:


src/Controller/OAuth/OAuthController.php
```diff
@@ -53,23 +53,25 @@ public function refreshRepoToken(string $type): Response
     /**
      * @param callable():string $emailProvider
      */
-    protected function createAndAuthenticateUser(callable $emailProvider, Request $request): Response
+    protected function createAndAuthenticateUser(string $type, callable $emailProvider, Request $request): Response
```
akondas, line 56: maybe `array $params = []` instead of string $type

karniv00l: we only need this when account doesn't exist (`if (!$this->guard->userExists($email)) {`) so we would have to unset this array in if statement 🤔

akondas: actually ... you're right

PR #100, opened 2020-04-23
Composer V2 Support

sadortun:
Added support for Composer V2.  Currently testing....

Todo
- [x] Add support for  `metadata-url`
- [ ] Add support for  `providers-api` (optional)

**V2 is just incretibly faster ! Its amaizing !** 

```bash
composer self-update --snapshot
composer clear-cache
time composer update
real    0m57.831s
user    0m0.030s
sys     0m0.060s

composer self-update --stable
composer clear-cache
time composer update
real    3m46.648s
user    0m0.015s
sys     0m0.075s
```


sadortun:
@akondas  Done plenty of tests in about 10 differents projects, for a total of 1600 installs.

I think this is ready to get merged.

akondas:
I'm very happy about this. And thank you very much for your time spent on work.

akondas:
Could you enable the "Allow edits from maintainers." option. I would add tests and we could merge.

sadortun:
@akondas No clue where is the option  `Allow edits from maintainers.`,  it should be in the sidebar, but its not ! I added you as a maintainer of my fork instead, hehhhe!

config/services.yaml
```diff
@@ -5,6 +5,7 @@
 # https://symfony.com/doc/current/best_practices/configuration.html#application-related-configuration
 parameters:
     package_name_pattern: '[A-Za-z0-9_.-]+/[A-Za-z0-9_./-]+?'
+    package_name_pattern_v2: '[A-Za-z0-9_.-]+/[A-Za-z0-9_./~-]+?'
```
akondas, line 8: why `~`?

sadortun: You need to be able to pass through dev packages, and they contain a `~`

https://packagist.org/p2/symfony/polyfill-php72~dev.json

sadortun: I guess you could merge it in the previous pattern, but i wanted to make it obvious

akondas: yes, we can combine it

src/Service/Proxy.php
```diff
@@ -61,6 +61,23 @@ public function providerData(string $package, int $expireTime = self::PACKAGES_E
...
+        if (!($fromPath = $this->metadataProvider->fromPath($package, $this->url, $expireTime))->isEmpty()) {
+            return $fromPath;
+        }
+
+        $providerPath = $this->getProviderPathV2($package);
+        if ($providerPath->isEmpty()) {
+            return Option::none();
+        }
+
+        return $this->metadataProvider->fromUrl($this->getUrl($providerPath->get()));
```
akondas, line 78: We will probably need to do it differently, in this case we will overwrite the v1 data with the v2 version.
If I find a moment, I can finish it, but currently I am still involved in something else, if you would like to finish it yourself, I suggest you do it through an additional parameter and save the data in a separate directory `p2`

akondas:
Thanks @sadortun for help. @karniv00l created complete support for v2 based on your work. I will add your name to changelog. We didn't want to do it in this PR so as not to waste your time. If we can do something more, let me know. Thanks again :clap:

Issue #85, opened 2020-04-22
GitLab package list limited to 40


iainfogg:
I've got access to a lot more than 40 repos in GitLab, but only 40 appear in the packages list when trying to add a new package. I assume this is because of an issue with the pagination in the GitLab API, but because there's no way to manually add a repo, I'm unable to add any repos that don't appear in those 40.

akondas:
Currently the paging mechanism is added and should download all repositories. But maybe we fetch them in a bad way. We are using `m4tthumphrey/php-gitlab-api` with this options:
```
$this->pager->fetchAll($this->client->projects(), 'all', [[
            'simple' => true,
            'owned' => true,
            'membership' => true,
            'order_by' => 'path',
        ]]))
```
flag desc:
```
@var bool   $owned       Limit by projects owned by the current user.
@var bool   $membership  Limit by projects that the current user is a member of.
```



Maybe some specific type of project is missing?

akondas:
Hello @iainfogg I checked the api and everything looks correct, we fetch a list of all projects, even if they do not fit on one page (which is not the best either).

Have you tried entering the name of the repository you were looking for? Maybe just not everything is displayed in this html component (or there may be a problem with js).

iainfogg:
Hi, yeah I've tried entering the full name of a package that's not in the 40, but it doesn't appear. if I examine the HTML in the page, the one I entered isn't in there either.

In the dropdown, I can't make any particular sense of the order - it's not in package, group or date order, so that doesn't help with any diagnostics I'm afraid.

akondas:
Thanks for the information. It seems to me that I am able to reproduce this situation. The problem is probably with workspace projects in which the user is not the owner.

akondas:
Hey, @iainfogg can you check please again and possibly close the issue if the problem is resolved?

iainfogg:
Hi @akondas - it's better now, but still not fully there. I have maybe 90 repos in GitLab, organised into subgroups. Originally when I raised the issue, only 3 of the approximately 60 repos in a subgroup (where all my packages are) were appearing. Now, if I type the name of the subgroup it shows 10 rather than 3.

Also, if I just type the name of a package which is in that group, but is not appearing in the list of 10, it does not appear either.

akondas:
[...]

PR #105, opened 2020-04-24
Fetch GitLab projects separately

karniv00l:


akondas:
This should help with #85

PR #106, opened 2020-04-24
Add ability to unlink OAuth integration from user profile page

karniv00l:
Screenshot 2020-04-27 at 13 25 41


src/Controller/Organization/PackageController.php
```diff
@@ -78,6 +79,7 @@ public function packageNew(Organization $organization, Request $request, GithubA
                 return $response;
             }
         } catch (HttpException $exception) {
+            $this->dispatchMessage(new RemoveOAuthToken($this->getUser()->id()->toString(), (string) $type));
```
akondas, line 82: I do not think this is a good idea, in the absence of a connection to the provider, we do not necessarily want to delete the token. I thought more about a separate action in the user profile.

PR #123, opened 2020-04-28
Return 404 when distribution file not found

akondas:


phpstan.neon
```diff
@@ -20,6 +20,10 @@ parameters:
             message: "#^Expression on left side of \\?\\? is not nullable\\.$#"
             count: 2
             path: src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
+        -
+            message: "#^Variable \\$http_response_header in isset\\(\\) always exists and is not nullable\\.$#"
+            count: 1
+            path: src/Service/Downloader/NativeDownloader.php
```
akondas, line 26: Reported, but "hard to solve" (https://github.com/phpstan/phpstan/issues/3213) so must ignore

src/Service/Dist/Storage/FileStorage.php
```diff
@@ -42,7 +43,9 @@ public function download(string $url, Dist $dist, array $headers = []): void
 
         AtomicFile::write(
             $filename,
-            $this->downloader->getContents($url, $headers)->getOrElseThrow(
+            $this->downloader->getContents($url, $headers, function () use ($url): void {
+                throw new NotFoundHttpException(sprintf('File not found at %s', $url));
```
akondas, line 47: this is quick and nasty fix, because this layer should not use this exception, but at the moment I don't see a better and faster way to fix this problem

PR #130, opened 2020-05-04
Fix doctrine proxy and UI for organization members

akondas:


phpstan.neon
```diff
@@ -1,6 +1,8 @@
 includes:
 	- vendor/phpstan/phpstan/conf/bleedingEdge.neon
 parameters:
+    excludes_analyse:
+        - src/Entity/Organization.php
```
akondas, line 5: I'm not happy with this, but I had to add default null because of https://github.com/doctrine/common/issues/886, so I exclude this file from analysis (I only hope temporarily). Another option is downgrade php ...

PR #145, opened 2020-05-07
Add package versions stats and tweak other charts

akondas:
Screenshot from 2020-05-07 08-46-24

### DIFF
--- a/src/DataFixtures/PrivatePackageDownloadFixtures.php
+++ b/src/DataFixtures/PrivatePackageDownloadFixtures.php
@@ -52,6 +52,7 @@ final class PrivatePackageDownloadFixtures extends Fixture
         $progress = new ProgressBar($output, 30);
         $output->writeln(sprintf('Package: %s', $package->name()));
         $progress->start();
+        $versions = array_map(fn () => $this->faker->numerify('#.#.#'), range(1, 10));
 
         for ($i = 0; $i < 30; ++$i) {
             $dayDownloads = random_int(0, 100);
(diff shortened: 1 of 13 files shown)


src/DataFixtures/PrivatePackageDownloadFixtures.php
```diff
@@ -52,6 +52,7 @@ private function loadDownloads(PackageName $package): void
         $progress = new ProgressBar($output, 30);
         $output->writeln(sprintf('Package: %s', $package->name()));
         $progress->start();
+        $versions = array_map(fn () => $this->faker->numerify('#.#.#'), range(1, 10));
```
akondas, line 55: there will be less granulation, otherwise it is after 500 versions which is rare ... or maybe not :thinking:

templates/base.html.twig
```diff
@@ -122,7 +122,7 @@
     {% endblock body %}
     {% include 'component/confirmationModal.html.twig' %}
 
-    <script src="https://code.jquery.com/jquery-3.4.1.slim.min.js" integrity="sha384-J6qa4849blE2+poT4WnyKhv5vZF5SrPo0iEjwBvKU7imGFAV0wwj1yYfoRSJoZ+n" crossorigin="anonymous"></script>
+    <script src="https://code.jquery.com/jquery-3.5.1.min.js" integrity="sha256-9/aliU8dGd2tb6OSsuzixeV4y/faTqgFtohetphbbj0=" crossorigin="anonymous"></script>
```
akondas, line 125: require for ajax call in version stats

templates/component/js/stats.js.twig
```diff
@@ -12,9 +12,6 @@
                     fontFamily: 'inherit',
                     height: 440,
                     parentHeightOffset: 0,
-                    toolbar: {
-                        show: false,
-                    },
```
akondas, line 17: it can be useful
Screenshot from 2020-05-07 09-22-45

PR #151, opened 2020-05-08
Fix: Download un-synced packages

jfcalcerrada:
Sorry if I'm not following the contributions guide, fell free to take it from here.


akondas:
Thanks for the contribution. One of Repman's main goals is to be a packagist proxy. For this we only download packages that are used by your organization. What's the point of downloading everything? Then it would be more of a mirror, and at the moment there are ready solutions that allow you to make a packagist mirror.

jfcalcerrada:
🤦  I was having some troubles setting up the Proxy, so I went into debugging some bits.
I got this completely the all way around.

PR #152, opened 2020-05-11
Allow user to disable account registration

### LINKED ISSUE #113: [Feature request] Ability to disable registration

karniv00l:
Resolves #113


src/Service/Symfony/RegistrationRouteListener.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Buddy\Repman\Service\Symfony;
+
+use Buddy\Repman\Service\Config;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+final class RegistrationRouteListener implements EventSubscriberInterface
```
akondas, line 13: it seems to me unnecessary, an additional `if` in the registration controller would suffice, instead of checking it every time for each request

src/Service/Config.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public function invalidate(): void
+    {
+        $this->cache->delete(self::CACHE_KEY);
+    }
+
+    private function load(): void
+    {
+        $this->values = $this->cache->get(
```
akondas, line 57: maybe additional check `if($values === [])`

Issue #150, opened 2020-05-08
Security scanner


akondas:
This feature will allow you to scan (`sensiolabs/security-checker`) downloaded packages and display the scan results in the package list. 

- [ ] the latest package will always be scanned
- [ ] it will be possible to run the scanner again for the selected package
- [ ] each package will have a scan history
- [ ] if possible we will add a warning for composer to display a relevant warning for the end user

PR #157, opened 2020-05-15
Implement security scanner

karniv00l:
Screenshot 2020-05-14 at 20 19 54
Screenshot 2020-05-14 at 20 18 35
Screenshot 2020-05-14 at 20 18 01
Screenshot 2020-05-14 at 20 09 34
Screenshot 2020-05-14 at 14 08 29

#150


src/Service/Organization/PackageManager.php
```diff
@@ -11,7 +11,7 @@
 use Munus\Control\Option;
 use Symfony\Component\Filesystem\Filesystem;
 
-final class PackageManager
+class PackageManager
```
karniv00l, line 14: I'm mocking this class in unit tests

composer.json
```diff
@@ -72,7 +73,6 @@
         "phpstan/phpstan-strict-rules": "^0.12.2",
         "phpstan/phpstan-symfony": "^0.12.3",
         "phpunit/phpunit": "^8.5",
-        "sensiolabs/security-checker": "^6.0",
```
karniv00l, line 75: moved from dev

config/packages/messenger.yaml
```diff
@@ -20,3 +20,4 @@ framework:
             'Buddy\Repman\Message\User\SendPasswordResetLink': async
             'Buddy\Repman\Message\User\SendConfirmToken': async
             'Buddy\Repman\Message\Proxy\AddDownloads': async
+            'Buddy\Repman\Message\Organization\Package\ScanPackage': async
```
akondas, line 23: Maybe `Buddy\Repman\Message\Security\ScanPackage` :thinking:

src/Controller/OrganizationController.php
```diff
@@ -321,6 +322,30 @@ public function stats(Organization $organization, Request $request): Response
...
+
+    /**
+     * @Route("/organization/{organization}/package/{package}/scan-results", name="organization_package_scan_results", methods={"GET","POST"}, requirements={"organization"="%organization_pattern%","package"="%uuid_pattern%"})
+     */
+    public function packageScanResults(Organization $organization, Package $package, Request $request): Response
+    {
+        return $this->render('organization/package/scanResults.html.twig', [
+            'organization' => $organization,
+            'package' => $package,
+            'results' => $this->packageQuery->getScanResults($package->id()),
```
akondas, line 345: I think we can add pagination right away

src/MessageHandler/Organization/SynchronizePackageHandler.php
```diff
@@ -30,5 +34,9 @@ public function __invoke(SynchronizePackage $message): void
         }
 
         $this->synchronizer->synchronize($package);
+
+        if ($package->isSynchronizedSuccessfully()) {
+            $this->messageBus->dispatch(new ScanPackage($message->id()));
+        }
```
akondas, line 40: we can leave it this way, but ultimately we could assume that we want to handle it later through the event, it's a pity that the unsuccessful publication of messages on the queue would affect the synchronization status (and this will happen now)

Alternatively, you could temporarily fix it this way: https://symfony.com/doc/current/messenger/message-recorder.html

src/Query/User/Model/ScanResult.php
```diff
@@ -0,0 +1,94 @@
...
+final class ScanResult
+{
+    private \DateTimeImmutable $date;
+    private string $status = ScanResultEntity::STATUS_PENDING;
+    private string $version;
+
+    /**
+     * @var mixed[]
+     */
+    private array $content;
```
akondas, line 18: this content is a bit magical to me, we can't cover it with more types

src/Query/User/PackageQuery/DbalPackageQuery.php
```diff
@@ -29,10 +30,42 @@ public function findAll(string $organizationId, int $limit = 20, int $offset = 0
...
+                    ORDER BY date DESC
+                    LIMIT 1
+                ) scan_result_date,
+                (
+                    SELECT s.content
+                    FROM organization_package_scan_result s
+                    WHERE s.package_id = p.id
+                    ORDER BY date DESC
+                    LIMIT 1
+                ) scan_result_content
```
akondas, line 65: why not left join? then if these columns are not present then just nulls will be returned

src/Service/Organization/PackageManager.php
akondas: And that's what worries me, that is, this class is doing too much and/or the interface and default implementation are missing.

src/Service/PackageScanner/SensioLabPackageScanner.php
```diff
@@ -0,0 +1,151 @@
...
+    {
+        $zip = new \ZipArchive();
+        $result = $zip->open($distFilename);
+        if ($result !== true) {
+            throw new \RuntimeException("Error while opening ZIP file '$distFilename', code: $result");
+        }
+
+        $content = null;
+        for ($i = 0; $i < $zip->numFiles; ++$i) {
+            if (preg_match('/\/composer.lock$/', (string) $zip->getNameIndex($i)) === 1) {
```
akondas, line 123: Will this only work for files in the root directory? Because there may be more composer.lock files there

akondas:
Still lack information about CVE and maybe a link (if you can link to the threat). For example: 
Screenshot from 2020-05-16 08-02-35

src/Query/User/PackageQuery/DbalPackageQuery.php
karniv00l: It bugs me also but If we join, grouping by those columns or agreggation is required and one of those columns is json which can't be done (not comparable)

karniv00l:
I've totally forgot about CVEs and links

karniv00l:
Screenshot 2020-05-18 at 20 08 56
Screenshot 2020-05-18 at 20 09 04

config/services.yaml
```diff
@@ -112,4 +112,11 @@ services:
         arguments:
             $url: '%env(APP_GITLAB_API_URL)%'
 
+    Buddy\Repman\Service\Doctrine\PackageSynchronizedListener:
```
akondas, line 115: honestly, I think I prefer to call it explicitly in the handler than such a hidden listener

src/Query/User/PackageQuery/DbalPackageQuery.php
akondas: Well, probably an additional query will be better

src/Command/ScanAllPackagesCommand.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $this
+            ->setName('repman:security:scan-all')
+            ->setDescription('Scan all synchronized packages')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $list = $this->packageQuery->getAllSynchronized();
```
akondas, line 44: this list should be paginated

config/services.yaml
```diff
@@ -108,8 +117,4 @@ services:
         arguments:
             - '@Gitlab\Client'
 
-    Buddy\Repman\Service\GitLabApi\RestGitLabApi:
-        arguments:
-            $url: '%env(APP_GITLAB_API_URL)%'
-
```
karniv00l, line 114: I've only moved this to the top since I noticed that this is marked as a 'vendor' section

Issue #154, opened 2020-05-12
 /login not found + There are no commands defined in the "proxy:domain" namespace.


bash-stack:
after manual install , and migrations , at the proxy-setup step:

` There are no commands defined in the "proxy:domain" namespace.`

Also  the / page redirects to `/login` and this gives a 404

akondas:
`proxy:domain` is command from Symfony CLI: https://symfony.com/download. Setting up subdomains in this way is rather mainly for development purposes. Production should be a redirection with an wildcard (for example at dns level). You can read more in docs: https://repman.io/docs/standalone/#dns-records

As for the second problem, the question is how is the application served (nginx, local php server or something else)?

bash-stack:
apache+php7.4fpm behing nginx proxy

bash-stack:
found it  , the "normal" nginx config makes the index.php/ORIGINALURL redirection , 

so a .htaccess containing the following lines in the public dir helped 

```
RewriteEngine on
RewriteCond %{REQUEST_FILENAME} !-f
RewriteCond %{REQUEST_FILENAME} !-d
RewriteRule ^(.*)$ /index.php/$1 [NC,L,QSA]
```

bash-stack:
found it  , the "normal" nginx config makes the index.php/ORIGINALURL redirection , 

so a .htaccess containing the following lines in the public dir helped 

```
RewriteEngine on
RewriteCond %{REQUEST_FILENAME} !-f
RewriteCond %{REQUEST_FILENAME} !-d
RewriteRule ^(.*)$ /index.php/$1 [NC,L,QSA]
```

PR #163, opened 2020-05-19
Create .htaccess

bash-stack:
fixing https://github.com/repman-io/repman/issues/154


akondas:
Although we do not officially support the configuration for Apache, we can do it, for people wanting to set up their own instance will definitely be useful. Thanks.

PR #170, opened 2020-05-22
Implement PackageScanner service

karniv00l:


src/Command/ScanAllPackagesCommand.php
```diff
@@ -0,0 +1,67 @@
...
+
+        for ($offset = 0; $offset <= $count; $offset = ($offset + 1) * $limit) {
+            $list = $this->packageQuery->getAllSynchronized($limit, $offset);
+
+            foreach ($list as $item) {
+                $this->scanner->scan(
+                    $this->packageRepository->getById(Uuid::fromString($item->id()))
+                );
+                $progressBar->advance();
+            }
```
akondas, line 59: We need to call EntityManager::clear to reduce memory usage, and use transaction here

src/Query/User/Model/ScanResult/ResultContent.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $this->json = $json;
+    }
+
+    public function html(): string
+    {
+        return $this->formatContent();
+    }
+
+    public function htmlSimple(): string
```
akondas, line 25: I would like to rename this method and maybe we can remove html from readl model?
I know it simplifies, but it's not their responsibility to format html.

src/Query/User/Model/ScanResult/ResultContent/LockFileResult.php
```diff
@@ -0,0 +1,71 @@
...
+            $body[] = "<b>$dependencyName</b> ($version)<ul>$advisories</ul>";
+        }
+
+        return implode('', $body);
+    }
+
+    /**
+     * @param Advisory[] $advisories
+     */
+    private function advisoriesHtml(array $advisories): string
```
akondas, line 47: same as above

src/Query/User/PackageQuery/DbalPackageQuery.php
```diff
@@ -29,10 +30,37 @@ public function findAll(string $organizationId, int $limit = 20, int $offset = 0
...
+                (
+                    SELECT status FROM organization_package_scan_result s
+                    WHERE s.package_id = p.id
+                    ORDER BY date DESC LIMIT 1
+                ) scan_result_status,
+                (
+                    SELECT content FROM organization_package_scan_result s
+                    WHERE s.package_id = p.id
+                    ORDER BY date DESC LIMIT 1
+                ) scan_result_content
```
akondas, line 59: we need to rewrite this query, fortunately, we have already agreed on a better solution

PR #171, opened 2020-05-22
Implement SecurityChecker

karniv00l:


config/services_test.yaml
```diff
@@ -1,6 +1,8 @@
 parameters:
     dists_dir: '%kernel.project_dir%/tests/Resources'
     repo_dir: '%kernel.project_dir%/tests/Resources'
+    security_advisories_db_dir: '%kernel.project_dir%/tests/Resources/fixtures/security/security-advisories'
+    security_advisories_db_repo: 'bogus'
```
akondas, line 5: `bogus`? :joy:

src/Service/Security/SecurityChecker/SensioLabsSecurityChecker.php
```diff
@@ -5,20 +5,210 @@
...
+    {
+        $matches = [];
+        preg_match('~^/(?<name>.+/.+)/.+yaml$~', $path, $matches);
+
+        return $matches['name'] ?? null;
+    }
+
+    private function loadAdvisoriesDatabase(): void
+    {
+        $this->advisories = $this->getAdvisories();
```
akondas, line 185: maybe we should check if is already loaded? (to prevent load twice in one request)

tests/Doubles/FakeProcess.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Buddy\Repman\Tests\Doubles;
+
+final class FakeProcess
```
akondas, line 7: it looks like an unused class

tests/Unit/Service/Security/SensioLabsSecurityCheckerTest.php
```diff
@@ -0,0 +1,134 @@
...
+    private string $dbDir;
+    private string $fixturesDir;
+
+    protected function setUp(): void
+    {
+        $this->dbDir = sys_get_temp_dir().'/repman/security-advisories';
+        $this->fixturesDir = __DIR__.'/../../../Resources/fixtures/security/locks';
+
+        $filesystem = new Filesystem();
+        $filesystem->mirror(
```
akondas, line 24: do you have to make a copy of it? Can't point the path to these fixture directly?
```php
new SensioLabsSecurityChecker(__DIR__.'/../../../Resources/fixtures/security/security-advisories', 'bogus');
```

tests/Unit/Service/Security/SensioLabsSecurityCheckerTest.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public function testSuccessfulScanWithoutAlerts(): void
+    {
+        self::assertEquals($this->checker->check($this->safeLock()), []);
+    }
+
+    public function testUpdateWithGitClone(): void
+    {
+        $this->expectException(ProcessFailedException::class);
```
akondas, line 112: why failed?

karniv00l: do we want tests to actually connect and fetch git repository?

tests/Unit/Service/Security/SensioLabsSecurityCheckerTest.php
karniv00l: during tests `.git` directory is created which may be left over in case of a test failure, interruption etc, this ensures that is always the same in the beginning

tests/Unit/Service/Security/SensioLabsSecurityCheckerTest.php
akondas: I do not think so

PR #177, opened 2020-05-26
Code review fixes

karniv00l:


src/Command/UpdateAdvisoriesDbCommand.php
```diff
@@ -34,7 +34,6 @@ protected function configure()
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $this->checker->update();
-        $output->writeln(sprintf('Database successfully updated'));
```
karniv00l, line 37: I don't think we need output from this command polluting cron logs 🤔

akondas: I agree

src/Entity/Organization/Package.php
```diff
@@ -200,6 +217,16 @@ public function latestReleasedVersion(): ?string
         return $this->latestReleasedVersion;
     }
 
+    /**
+     * @param mixed[] $result
+     */
+    public function setScanResult(string $status, \DateTimeImmutable $date, array $result): void
```
akondas, line 223: maybe: `packageScanned` :thinking: but it can stay like this now

PR #178, opened 2020-05-26
Update db command

karniv00l:


src/Service/Security/SecurityChecker/SensioLabsSecurityChecker.php
```diff
@@ -198,19 +204,23 @@ private function updateRepo(): void
...
         $process = new Process($command, $this->databaseDir);
         $process->run();
 
         if (!$process->isSuccessful()) {
             throw new ProcessFailedException($process);
         }
+
+        return $process->getOutput();
     }
 }
```
karniv00l, line 226: unfortunately:

Screenshot 2020-05-26 at 19 44 22

src/Service/Security/SecurityChecker/SensioLabsSecurityChecker.php
```diff
@@ -39,6 +40,11 @@ public function update(): void
         $this->updateRepo();
     }
 
+    public function hasDbBeenUpdated(): bool
+    {
+        return $this->dbUpdated;
```
akondas, line 45: please no, service should not have a state, better return object or bool from `update` method

PR #180, opened 2020-05-27
Make the first user an admin

nickygerritsen:
Currently the only way to create the first admin user is to update the record in the DB. This fixes it by making the first user an admin.

Two questions:
* I checked the existing (integration) test but it didn't test anything related to the roles, so I'm not sure if/how we can extend it. We could test in testSuccessfulRegistration that the body contains Administrator and in a separate test where we create a user with a fixture, that it does not?
* I'm not sure if putting the logic in the controller is the way to go. I tried looking at other places where this kind of logic resides but couldn't really find any.


nickygerritsen:
I totally missed that there is an `repman:create:admin` command. Still this might be useful?

akondas:
Sorry, but it doesn't seem like a good idea to me. We can optionally add a description to the documentation, or call this command during the first installation

For the future, you can always open the issue and talk things over so that you don't have to produce the code :wink: 

Thanks for your time :tropical_drink:

Issue #189, opened 2020-06-02
Missing composer.lock reported as error


mbrodala:
When adding a package without `composer.lock` Repman reports this as error:

![image](https://user-images.githubusercontent.com/5037116/83540064-1a66a280-a4f8-11ea-96a8-a8480258f2b2.png)

However, having a repository without `composer.lock` is perfectly valid, for example for library packages as well as avoiding issues with dependency overrides on CI build.

So IMO this should not be reported as error.

karniv00l:
I agree. @akondas any thoughts?

akondas:
Yes, package without `composer.lock` should not have `error` status. Maybe `missing` or `not applicable`? I also wonder if we can do the scan in any other way then?

akondas:
Fixed in #191 
![Screenshot from 2020-06-05 10-24-25](https://user-images.githubusercontent.com/8239917/83854352-e2897600-a716-11ea-8bf7-bbc66da61127.png)

mbrodala:
Thanks, after a re-scan the updated status is show. :+1:

PR #191, opened 2020-06-03
Send scan result to owners; Don't trigger error on missing lockfile

karniv00l:


templates/emails/scan-result.html.twig
```diff
@@ -0,0 +1,48 @@
...
+                    </table>
+                    <p style="font-size: 16px; line-height: 1.625; color: #51545E; margin: .4em 0 1.1875em;">
+                        Thanks,
+                        <br />The Repman Team
+                    </p>
+                </div>
+            </td>
+        </tr>
+    </table>
+{% endblock %}
```
karniv00l, line 48: Screenshot 2020-06-03 at 21 40 05

tests/Unit/Service/Security/SensioLabsPackageScannerTest.php
```diff
@@ -128,10 +131,16 @@ private function prepareScanner(string $fixtureType = 'repman'): SensioLabsPacka
 
         $packageManager->method('distFilename')->willReturn(Option::some($distFile));
 
+        $messageBusMock = $this->createMock(MessageBus::class);
+        $messageBusMock
+            ->method('dispatch')
+            ->willReturn(new Envelope(new SendScanResult('buddy', 'test/test', 'test', [])));
```
akondas, line 137: `willReturn` is unused in this case, better you some kind of assertion: `expects($this->once())`

src/MessageHandler/Security/SendScanResultHandler.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        $this->mailer = $mailer;
+        $this->organizationQuery = $organizationQuery;
+    }
+
+    public function __invoke(SendScanResult $message): void
+    {
+        $organization = $this->organizationQuery
+            ->getByAlias($message->organizationAlias())
+            ->getOrElseThrow(new \RuntimeException('Organization not found'));
```
akondas, line 27: we won't do anything about it anyway, it may be better to just ignore this mistake in this case and just return

src/Message/Security/SendScanResult.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * @var mixed[]
+     */
+    private array $result;
+
+    /**
+     * @param mixed[] $result
+     */
+    public function __construct(string $organizationAlias, string $packageName, string $packageId, array $result)
```
akondas, line 21: Maybe we can add `emails` array to this message. This will simplify the handler. You can easliy add `ownersEmails()` method to `Package` entity

src/Service/Security/PackageScanner/SensioLabsPackageScanner.php
```diff
@@ -69,18 +77,17 @@ private function saveResult(Package $package, string $status, array $result): vo
...
-    /**
-     * @param array<string,array<string,string>> $error
-     */
-    private function saveError(Package $package, array $error): void
-    {
-        $status = ScanResult::STATUS_ERROR;
-        $date = new \DateTimeImmutable();
-        $package->setScanResult(ScanResult::STATUS_ERROR, $date, $error);
-        $this->results->add(new ScanResult(Uuid::uuid4(), $package, $date, $status, $error));
+        $this->messageBus->dispatch(new SendScanResult(
```
akondas, line 85: I don't think that an email should be sent every time. It would be better to do it only for errors - in that case when security vulnerability occurs.

src/Service/Security/PackageScanner/SensioLabsPackageScanner.php
```diff
@@ -144,10 +151,6 @@ private function extractLockFiles(string $distFilename): array
 
         $zip->close();
 
-        if ($lockFiles === []) {
-            throw new \RuntimeException('Lock file not found');
-        }
```
akondas, line 149: :thinking: what about new status? see https://github.com/repman-io/repman/issues/189#issuecomment-637975628

tests/Unit/Service/Security/SensioLabsPackageScannerTest.php
karniv00l: this is actually done to supress warnings (Envelope cannot be mocked) since dispatch() method must return Envelope

src/Service/Security/PackageScanner/SensioLabsPackageScanner.php
karniv00l: You're right, I've added `n/a` status

src/MessageHandler/Security/SendScanResultHandler.php
```diff
@@ -0,0 +1,33 @@
...
+
+final class SendScanResultHandler implements MessageHandlerInterface
+{
+    private Mailer $mailer;
+    private OrganizationQuery $organizationQuery;
+
+    public function __construct(Mailer $mailer, OrganizationQuery $organizationQuery)
+    {
+        $this->mailer = $mailer;
+        $this->organizationQuery = $organizationQuery;
```
akondas, line 20: not used :scissors:

src/Service/Mailer/SymfonyMailer.php
```diff
@@ -64,4 +64,26 @@ public function sendInvitationToOrganization(string $email, string $token, strin
...
         );
     }
+
+    /**
+     * @param string[] $emails
+     * @param mixed[]  $result
+     */
+    public function sendScanResult(array $emails, string $packageName, string $packageId, string $organizationAlias, array $result): void
+    {
+        foreach ($emails as $email) {
```
akondas, line 74: we can remove `foreach` and use `setBcc` to send one email to multiple receivers

src/Entity/Organization/Package.php
```diff
@@ -205,14 +205,10 @@ public function webhookWasRemoved(): void
...
-            if ($member->isOwner()) {
-                $emails[] = $member->email();
-            }
-        }
-
-        return $emails;
+        return $this->organization->members()
+            ->filter(fn ($member) => $member->isOwner())
+            ->map(fn ($member) => $member->email())
+            ->toArray();
```
akondas, line 211: :rocket:

PR #196, opened 2020-06-05
Send email in separate requests since bcc not work for aws ses :angry:

akondas:


src/Service/Mailer/SymfonyMailer.php
```diff
@@ -71,17 +71,19 @@ public function sendInvitationToOrganization(string $email, string $token, strin
...
-            ->subject("Vulnerabilities found in $packageName package")
-            ->htmlTemplate('emails/scan-result.html.twig')
-            ->context([
-                'packageName' => $packageName,
-                'packageId' => $packageId,
-                'organizationAlias' => $organizationAlias,
-                'result' => $result,
-            ])
-        );
+        foreach ($emails as $email) {
```
akondas, line 74: Sorry @karniv00l, I wanted well, it came out as always :facepalm:

Issue #159, opened 2020-05-19
Create regular user from CLI (command)


bash-stack:
#### Circumstances:
* custom gitlab
* disabled user registration to prevent spam

→ currently creating new users from cli as admin and dropping their rights  , unsuitable for serious deployments

## suggestions: 
- [x] add `bin/console  repman:create:user` command to add regular user
- [ ] add web-gui option "`only allow account creation from` gitlab/github/providerXYZ"  when there are ONLY gitlab/github/providerXYZ users

akondas:
First looks ok. Second option need more detailed description, I think it will be related with (#58). Generally, such integration will open a window for even more possibilities, so it will definitely jump on our roadmap. Thanks for the feedback.

bash-stack:
The second one should go like this:

frontend with only ONE auth button -> gitlab/github/bitbucket -> create user 

since there should be no public registration , only oauth'ies from  self hosted gitlab/younameit should be able to register

nickygerritsen:
For the second part of this issue: the problem currently is that disabling registration disables both local and oauth registration.

Perhaps a new option, 'disable local authentication' would fix this? This would then disabling:
* Locally logging in with a username/password
* Locally registering
* Changing your password

I could create a PR for that if that sounds like a good idea? It is something we would like to have, since we use GitLab for all our authentication, so we don't want/need local users login all.

akondas:
Sounds good, we can do separate options with this:
- [ ] User email/password registration
- [ ] User oath registration

nickygerritsen:
Don’t we have oauth registration already because we only show oauth buttons for env vars you have set?

akondas:
Yes and no. Env vars show and hide buttons for both registration and login, this is a different concept. This option in the administration allows you to directly enable / disable the registration process itself.

nickygerritsen:
Ok but I was proposing also hiding *login* for email/password users. So a bit like the env var.
Then maybe split it like this:
* checkbox for oauth reg
* select for email/password with three options: allow login+reg, allow login but disable reg; fully disallow

We don’t have the select for oauth since we have the env var to do that.

Thoughts?

akondas:
It sounds ok to me. @karniv00l?

bash-stack:
(reminder: currently the oauth providers have to be defined , but empty , so 
`OAUTH_EXAMPLE-PLAFTORM_VALUETHINGY=` works , but not defining kills the system )

karniv00l:
Looks good to me

PR #200, opened 2020-06-14
Registration config options.

nickygerritsen:
This should be what's discussed in #159 .
I also added default empty strings to all OAuth environment variables, reported by @bash-stack in that same issue.

I do wonder what we should do with the existing configuration option: I now just delete it completely and insert new values, but we could also automatically set the values of the new data based on the old value with a 'fancy' query. Then the question is what about the `down` method if the two new values don't 'agree'?

### DIFF
--- a/src/Migrations/Version20200614133313.php
+++ b/src/Migrations/Version20200614133313.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Buddy\Repman\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20200614133313 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql("INSERT INTO config (key, value)
+VALUES ('local_authentication', (
+    SELECT CASE
+               WHEN value = 'enabled'
+                   THEN 'login_and_registration'
+               ELSE 'login_only'
+               END
+    FROM config
+    WHERE key = 'user_registration'
+))");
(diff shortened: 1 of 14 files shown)


nickygerritsen:
(checking the failed tests)

nickygerritsen:
Ah I misunderstood the `env(default:X:Y)` option and I think it's not possible to not have to specify the variables, unless someone thinks otherwise?

nickygerritsen:
I removed the second commit but it seems Buddy doesn't pick that up?

akondas:
Fancy query will always cause more confusion, migration is a good idea. It seems ok to me, if @karniv00l approve, then we can merge this.

src/Migrations/Version20200614133313.php
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');
+
+        $this->addSql("INSERT INTO config (key, value) VALUES ('local_authentication', 'login_and_registration')");
+        $this->addSql("INSERT INTO config (key, value) VALUES ('oauth_registration', 'enabled')");
+        $this->addSql("DELETE FROM config WHERE key = 'user_registration'");
```
akondas, line 27: What about instances where this option was already used, we need to consider this in migration and set:
- [ ] if `user_registration` was enabled then leave as now
- [ ] if `user_registration` was disabled then `user_registration` = `disabled` and `local_authentication` = `login_only`

correct me if I'm wrong

nickygerritsen: That was my point. Should we try to gracefully upgrade? And what about downgrade?

akondas:
The rest works fine, I tested locally and everything is ok. Thanks @nickygerritsen :+1:

src/Migrations/Version20200614133313.php
akondas: What you mean by `gracefully upgrade`?
In `down` migration we check `local_authentication` if is `login_only` and the set `user_registration` to `disabled` (otherwise `enabled`

nickygerritsen: OK that makes sense. I will fix that

nickygerritsen:
Something like this? Not sure if there is any policy on the multiline SQL statements

akondas:
Thanks @nickygerritsen

Issue #96, opened 2020-04-23
Allow for public (read-only) access in organization


akondas:
This issue is a continuation of the thread from reddit:
> I mean adding a package to an organisation, and guests being able to browse and download those packages.

We could do an option in your organization's settings that would allow open access for added packages.

sadortun:
I am not sure if this would fit in this issue but i can see 3 different access levels to access the organizations.

- Authenticated admin ( add, remove packages, add users .... )
- Authenticated users ( can only access packages )
- Anonymous  users (No authentification required to access the organizations package in read-only)

For the Anonymous access, it could probably be a checkbox `Allow Anonymous users` in the admin settings

akondas:
These levels of access are already partially implemented by members of the organization. Only the option with anonymous access to the organization remains. I think we will do it in the next release (right after security scanner).

akondas:
Released in https://github.com/repman-io/repman/releases/tag/0.4.0

PR #201, opened 2020-06-18
Implement anonymous access to organization

karniv00l:
As requested in #96

### DIFF
--- a/src/Form/Type/Organization/ChangeAnonymousAccessType.php
+++ b/src/Form/Type/Organization/ChangeAnonymousAccessType.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Buddy\Repman\Form\Type\Organization;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+class ChangeAnonymousAccessType extends AbstractType
+{
+    public function getBlockPrefix(): string
+    {
+        return '';
+    }
+
+    /**
+     * @param array<mixed> $options
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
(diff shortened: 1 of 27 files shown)


karniv00l:
Screenshot 2020-06-18 at 16 12 03
Screenshot 2020-06-18 at 16 11 41
Screenshot 2020-06-18 at 16 11 22

src/Form/Type/Organization/ChangeAnonymousAccessType.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * @param array<mixed> $options
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('hasAnonymousAccess', CheckboxType::class, [
+                'label' => 'Allow anonymous users',
+                'required' => false,
```
akondas, line 27: please add a "hint" that enabling anonymous access allows to browse and download all organization packages

src/Security/AnonymousUserAuthenticator.php
```diff
@@ -0,0 +1,77 @@
...
+        if ($organizationAlias === null) {
+            throw new BadCredentialsException();
+        }
+
+        return $organizationAlias;
+    }
+
+    public function getUser($credentials, UserProviderInterface $userProvider)
+    {
+        return $userProvider->loadUserByUsername('alias:'.$credentials);
```
akondas, line 50: why not check if `$userProvider` is `OrganizationProvider` and use `loadUserByAlias` method (as public). This will allow to remove this `alias:` magic string.

karniv00l: That was also my first thought, but it is not a part of `UserProviderInterface` so phpstan will be unhappy about undeclared method

akondas: `instance of` will help with that

src/Form/Type/Organization/ChangeAnonymousAccessType.php
karniv00l: Screenshot 2020-06-19 at 11 20 26

Issue #175, opened 2020-05-23
Package View Improvements


giggsey:
- [x] A copy button for the package name (makes installation much easier)
 - [x] Ability to view the README for a package
 - [x] List of available releases
 - [ ] List of dependencies

Mimics our commonly used features from satis/packagist

nickygerritsen:
I wanted to look into implementing this. Some questions / remarks:

* For the copy button I guess we can use the existing clipboard.svg and add a small button after the name?
* For the readme, available releases and dependencies, see below

For the readme: I think this is the hardest one, since we need to actually get the readme; this is not part of the composer `PackageInterface`. First of all, it's probably best to use the latest release, since we have more logic for that already. Then what we *could* do is use the existing `PackageManager::distFilename` function to find the zip/tar for that version and then check the contents. However:

* How to handle tar files? I saw the recent SensioLabsPackageScanner only handles zip files so should we also do that here?
* How to determine the filename of the readme file? We can have `readme` and `readme.md` (and with caps) and probably some more, so we should determine an order. Also, it could happen there is a readme in a subfolder. How do we handle that we only want the root one? My zip files have one root folder and then in there are the files, but I haven't found the logic of how that root folder is named yet. Then if it is markdown we can use https://github.com/KnpLabs/KnpMarkdownBundle to render it?
* Where do we store the readme? Add a field to `Buddy\Repman\Entity\Organization\Package`?

For releases and dependencies, as well as the readme, I propose to add a `OrganizationController::package` action for the GET URL `/organization/{organization}/package/{package}` which can display all this info. Getting the versions and dependencies we can do by calling `PackageManager::findProviders` although I wonder if we should add a `findProvider` that takes one package?
Packagist shows links to required packages that also exist, we could do that if we add something like `PackageRepository::findByNames`? I'm not sure how you normally handle if you give it 5 names and only 3 exist. I only see that in `DbalDownloadsQuery::findByNames` but that's a bit different, it's not in a repository.

To make life even easier we could show a copy button after each version on this detail page, that will copy `"my/package": "1.2.3"` to the clipboard?

Any thoughts? We could also start with the simpler ones, like the copy button and the releases/dependencies.

giggsey:
> I'm not sure how you normally handle if you give it 5 names and only 3 exist. 

I think linking through to packagist for those non-local packages would be fine.

> To make life even easier we could show a copy button after each version on this detail page, that will copy "my/package": "1.2.3" to the clipboard?

[...]

PR #208, opened 2020-07-04
Add support for showing all package versions. Refs #175.

nickygerritsen:
In #175 there was a request to show the package versions. Since that is also a very useful feature for us, I have added this functionality.

Some notes / remarks:
* I added a package details page so one can view the versions here. We could later possibly add more here (like the readme). I made anonymous organization users have access to it.
* You can get to the details page both using the dropdown menu as well as clicking the package name.
* Does showing the clipboard button make sense? For `dev-*` versions this is what you want but for other versions you might want to add `^` or `~` in front of it. Maybe only let one copy the version itself as well as a separate button to copy the package name?
* I'm not sure if my logic in the Package and Version entity make sense?
* Perhaps we should add a `stable` boolean to version so we can show the `dev-*` versions in a separate table?
* I modified some tests to also test if we have the correct versions, but I'm not sure how/if we should test the reference as well? We could do it for the artifact case, since these ZIP's should have versions. For the local path one we can't assume anything since it uses local git info.

Some screenshots:
image
image

### DIFF
--- a/src/Entity/Organization/Package.php
+++ b/src/Entity/Organization/Package.php
@@ -5,6 +5,9 @@ declare(strict_types=1);
 namespace Buddy\Repman\Entity\Organization;
 
 use Buddy\Repman\Entity\Organization;
+use Buddy\Repman\Entity\Organization\Package\Version;
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping as ORM;
 use Ramsey\Uuid\UuidInterface;
 
@@ -101,6 +104,12 @@ class Package
     private ?\DateTimeImmutable $lastScanDate = null;
 
     /**
+     * @var Collection<int,Version>|Version[]
+     * @ORM\OneToMany(targetEntity="Buddy\Repman\Entity\Organization\Package\Version", mappedBy="package", cascade={"persist"}, orphanRemoval=true)
+     */
+    private Collection $versions;
+
+    /**
      * @param mixed[] $metadata
      */
     public function __construct(UuidInterface $id, string $type, string $url, array $metadata = [])
@@ -109,6 +118,7 @@ class Package
         $this->type = $type;
         $this->repositoryUrl = $url;
         $this->metadata = $metadata;
+        $this->versions = new ArrayCollection();
     }
 
     public function id(): UuidInterface
@@ -134,12 +144,20 @@ class Package
         return $this->repositoryUrl;
     }
 
-    public function syncSuccess(string $name, string $description, string $latestReleasedVersion, \DateTimeImmutable $latestReleaseDate): void
+    /**
+     * @param string[] $encounteredVersions
+     */
+    public function syncSuccess(string $name, string $description, string $latestReleasedVersion, array $encounteredVersions, \DateTimeImmutable $latestReleaseDate): void
     {
         $this->setName($name);
         $this->description = $description;
         $this->latestReleasedVersion = $latestReleasedVersion;
         $this->latestReleaseDate = $latestReleaseDate;
+        foreach ($this->versions as $version) {
+            if (!in_array($version->version(), $encounteredVersions, true)) {
+                $this->versions->removeElement($version);
+            }
+        }
         $this->lastSyncAt = new \DateTimeImmutable();
         $this->lastSyncError = null;
     }
@@ -245,5 +263,34 @@ class Package
         }
 
         $this->name = $name;
+    }
+
+    /**
+     * @return Collection<int,Version>|Version[]
+     */
+    public function versions(): Collection
+    {
+        return $this->versions;
+    }
+
+    public function addOrUpdateVersion(Version $version): void
+    {
+        if ($this->getVersion($version->version()) !== false) {
+            $this->getVersion($version->version())->setReference($version->reference());
+            $this->getVersion($version->version())->setSize($version->size());
+
+            return;
+        }
+
+        $version->setPackage($this);
+        $this->versions->add($version);
+    }
+
+    /**
+     * @return Version|false
+     */
+    public function getVersion(string $versionString)
+    {
+        return $this->versions->filter(fn (Version $version) => $version->version() === $versionString)->first();
     }
 }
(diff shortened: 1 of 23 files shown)


nickygerritsen:
Not sure what I should do with the remaining phpstan errors, but those are related to one of my questions about how to handle this version logic

src/Query/User/PackageQuery/DbalPackageQuery.php
```diff
@@ -103,6 +104,29 @@ public function getById(string $id): Option
         return Option::some($this->hydratePackage($data));
     }
 
+    /**
+     * @return Version[]
+     */
+    public function getVersions(string $packageId): array
```
akondas, line 110: it seems to me that it should be a paged list, because the quantity may grow, but at the moment it can stay that way (we can do it in a separate PR)

templates/organization/packages.html.twig
```diff
@@ -16,7 +16,7 @@
         <thead>
             <tr>
                 <th>Name</th>
-                <th>Version</th>
+                <th>Latest version</th>
```
akondas, line 19: :+1:

src/Entity/Organization/Package.php
```diff
@@ -246,4 +256,44 @@ private function setName(string $name): void
...
+     */
+    public function getVersion(string $versionString)
+    {
+        return $this->versions->filter(fn (Version $version) => $version->version() === $versionString)->first();
+    }
+
+    /**
+     * @param string[] $encounteredVersions
+     */
+    public function removeUnencounteredVersions(array $encounteredVersions): void
```
akondas, line 291: what about put this logic to `syncSuccess` method? Now this method will be always called before `syncSuccess`, so i think we can add this `$encounteredVersions` param there.

templates/organization/package/details.html.twig
```diff
@@ -0,0 +1,69 @@
...
+
+        <h3>Available versions</h3>
+
+        <table class="table">
+            <thead>
+            <tr>
+                <th>Version</th>
+                <th>Released</th>
+                <th>Reference</th>
+                <th></th>
```
akondas, line 38: Maybe we can add `size` column here?

akondas:
Looks great to me, thanks for your time :beers:.

src/Entity/Organization/Package.php
nickygerritsen: Maybe even pass the whole packages to syncsuccess and move the adding/removing logic fully there?

templates/organization/package/details.html.twig
nickygerritsen: Any idea how I can get the size easily? I could add a ->size to `distStorage` that returns the size of the archive?

src/Entity/Organization/Package.php
akondas: make sense

templates/organization/package/details.html.twig
akondas: Sounds good: `Storage -> size(Dist $dist): int`

src/Entity/Organization/Package.php
nickygerritsen: So given the below idea about storage: either we need to pass $diskStorage to `syncSuccess` or not move the whole packages there. I think the second option is better? Since otherwise we are passing services.

nickygerritsen:
Also a colleague of my suggested that it might be more logical to add a copy button for the `composer require` command. Shall I do the following two things:
* Add a general copy button at the top that will `composer require <packagename>`; and
* Change the copy button for each such that it will `composer require <packagename>:<version>`

?

akondas:
> Also a colleague of my suggested that it might be more logical to add a copy button for the `composer require` command. Shall I do the following two things:
> 
> * Add a general copy button at the top that will `composer require <packagename>`; and
> * Change the copy button for each such that it will `composer require <packagename>:<version>`
> 
> ?

This also sounds good :+1:

nickygerritsen:
Fixes:
* Added pagination. Wasn't too hard so why not 🤷‍♂️ 
* Added size column
* Changed logic as requested, although adding/updating versions is still in the synchronize method, size we need the distStorage
* Updated the copy button things
* Added more tests to have at least the same coverage as before 🎉

src/Entity/Organization/Package.php
akondas: Passing these services is not bad. Generally, we will be doing a slight change of architecture towards a more layered (domain, application, infrastructure) to get away from this symfony style. At the moment it can stay as it is :+1:

akondas:
@nickygerritsen After testing, I suggest you do one more thing: maybe we can add info when no versions are available:
Screenshot from 2020-07-07 08-11-13

```
To see available versions of the package, please perform synchronization again.
````
and button `Sync now`

Because currently all packages that have not been synchronized will have this page blank. The rest works fine. Good job :beers:

akondas:
And what do you thing about something like this:
Screenshot from 2020-07-07 08-19-50
Light space compression. If it's better then I can add it to your PR.

nickygerritsen:
I will do the first thing then you can do the second, looks much better.

nickygerritsen:
image
Tada, like this?

akondas:
:facepalm: Actually, we don't really need this extra table, after all, all available versions can be fetch from the metadata file `var/repo/organization/p/vendor/package.json`. You do not need to change anything during synchronization. In general, it facilitates the whole process. Just write a parser that extracts this data from this file. Sorry, I don't know why I didn't think about it before ... :thinking:

nickygerritsen:
> 🤦 Actually, we don't really need this extra table, after all, all available versions can be fetch from the metadata file `var/repo/organization/p/vendor/package.json`. You do not need to change anything during synchronization. In general, it facilitates the whole process. Just write a parser that extracts this data from this file. Sorry, I don't know why I didn't think about it before ... 🤔

But will it be fast enough? Also pagination is then less useful since we need to read and `json_decode` the whole file anyway? Or do you still want it for client-side render speed?

I can try to write a parser for this data.

akondas:
Actually, that's right, I saw 30MB files, that would kill rendering this page. Let's stay with what you managed to do now. If this does not work, we can always change this :wink:

akondas:
This is my little UI refactor: https://github.com/Lets-Talk-NL/repman/pull/1/files :wink:

nickygerritsen:
Nice, merged and squashed

akondas:
Thanks @nickygerritsen

PR #209, opened 2020-07-05
Add support for using S3 as storage

jonnyynnoj:
This integrates [thephpleague/flysystem](https://github.com/thephpleague/flysystem) for proxy & organization storage. By default it uses the local adapter and stays compatible with the current storage implementation. It includes the S3 adapter which can be chosen by setting the correct env vars.

This is needed as we are looking to run it in GKE, but the GCEPersistentDisk persistent volume doesn't support ReadWriteMany and so cannot be used for shared storage.

The only real change is the way the security advisories are stored. Rather than using a cron to update them, they are now updated immediately before a single package scan or all packages scan.


akondas:
> The only real change is the way the security advisories are stored. Rather than using a cron to update them, they are now updated immediately before a single package scan or all packages scan.

It was done for a reason. We don't want to scan packages unnecessarily if the security database has not changed . Just as we don't want to download a new database before each scan. At the moment, this database is updated several times a week so there is no need to check it before each scan.

Working with this adapter is a good job, but removing security database update is unacceptable at the moment. We can possibly increase the frequency of this cron or wonder if this update can be carried out in any other way (but not before every scan).

jonnyynnoj:
Hey, thanks for the feedback.

> We don't want to scan packages unnecessarily if the security database has not changed

Good point 👍 . I can change this back so that it only scans packages if the database has changed.

> Just as we don't want to download a new database before each scan

Not sure I see the issue with this though? Considering the update is just a git pull, updating the database before a single package scan isn't going to cause much of a performance hit. Plus, it guarantees that new packages are scanned against the latest advisories rather than a database X hours out of date depending on the frequency of the cron update.

akondas:
This database doesn't change that often, and cron will refresh the database anyway and trigger a scan if needed. With a few packages it is not a problem. But if there are thousands of them, then why bomb this repository since I know that with high probability in X hours this database will not change. If a new vulnerability is detected, you still need to scan all packages again.

But for such matters I propose to do differently.
If you care so much about it, let's do it as an option in the configuration:
`Update security database before each scan` -> yes/no
With this help help message: `Vulnerability database is updated every 2 hours, enable this feature if you want to update database before each package scan`.

Default value will be `no` (so logic will stay like this). For your instance it will be possible to switch it.
I will gladly merge something like this :wink:

jonnyynnoj:
> But if there are thousands of them, then why bomb this repository

So for proxy packages it doesn't matter how many there are, right? There's 1 update and then all the packages get scanned.

For organization packages it will do an update before scanning each package, when either adding them or when something triggers a synchronization. But even with thousands of packages I'm not sure this really counts as "bombing" the repository. Especially since adding repositories will be mostly a one-time setup. And even then, it's just a git pull that will 99.9% of the time result in 'up to date' as the database rarely changes.

> If you care so much about it, let's do it as an option in the configuration:

Well the thing is that without being able to use a shared volume, this is more required than an option. Basically there are 2 options to to manage the security database:
1. Sync the git repository into S3.
2. Make the cron & consumer containers have their own copies of the database which they will both be able to update when needed.

I wanted to avoid option 1 since it would require more complex logic to keep the repository and S3 contents in sync. Instead I felt that doing a git pull before the updates would be a simpler and acceptable solution.

akondas:
This still doesn't exclude doing so as an option. Unfortunately for our public instance, I can't merge something that will permanently disable this update. Your cron and consumer containers can read this option and work both ways: with auto update enabled or disabled.

jonnyynnoj:
Could you explain a bit more why the auto update is needed? I feel like I must be missing something 😅  I can't see the difference between the cron updating in the background vs updating just before the scan.

akondas:
Our public instance has thousands of packages and is constantly growing. I see no sense to check before each scan if the database is up to date, since cron will check it every 2 hours. In addition, if after checking it turns out that there are new vulnerabilities, then all packages will be automatically scanned.

As I wrote earlier, I do not mind the proposed solution. I like it very much. The only condition is that it is a matter of instance configuration. Then everyone will personally consider whether they want to check the vulnerability database every time, whether it is enough for him every two hours.

akondas:
Something like this (yes/no configuration option):
Screenshot from 2020-07-07 13-15-15

PR #215, opened 2020-07-14
Fix user voters with anonymous access

karniv00l:
- Handle non-existing organization for anonymous user
- Fix problem with authenticated user trying to access public organization
- Handle organizations named `new*`


src/Service/Organization/OrganizationVoter.php
```diff
@@ -47,16 +47,21 @@ public function vote(TokenInterface $token, $subject, array $attributes): int
      */
     protected function voteOnAttribute(string $attribute, $subject, TokenInterface $token): bool
     {
+        /**
+         * @var User
+         */
         $user = $token->getUser();
-        if (!$user instanceof User) {
-            return false;
-        }
```
karniv00l, line 53: voter should abstain in this case

PR #222, opened 2020-07-16
Serve static proxy metadata and use v2 endpoint for dist lookup

akondas:


src/Controller/OrganizationController.php
```diff
@@ -342,14 +342,6 @@ public function packageScanResults(Organization $organization, Package $package,
         ]);
     }
 
-    protected function getUser(): User
```
akondas, line 345: dead code :skull_and_crossbones:

tests/Functional/Controller/ProxyControllerTest.php
```diff
@@ -46,27 +46,28 @@ public function testProviderAction(): void
                 "buddy-works/repman": "@array@"
             }
         }
-        ', $this->client->getResponse()->getContent());
+        ', $response);
+        self::assertTrue($this->client->getResponse()->isCacheable());
```
akondas, line 50: this will allow us to let others use their cache as well

PR #224, opened 2020-07-17
Implement sync proxy metadata command

akondas:


src/Service/Proxy.php
```diff
@@ -92,8 +92,9 @@ public function download(string $package, string $version): void
 
         foreach ($this->decodeMetadata($package) as $packageData) {
             $lastDist = $packageData['dist'] ?? $lastDist;
-            if ($version === $packageData['version']) {
-                $this->filesystem->write($this->distPath($package, $lastDist['reference'], $lastDist['type']), $this->downloader->getContents($lastDist['url'])
+            $path = $this->distPath($package, $lastDist['reference'], $lastDist['type']);
+            if ($version === $packageData['version'] && !$this->filesystem->has($path)) {
```
akondas, line 96: added check if version already exist (without this `$this->filesystem->write` will throw exception)

tests/Doubles/FakeDownloader.php
```diff
@@ -9,8 +9,14 @@
 
 final class FakeDownloader implements Downloader
 {
+    // todo: remove this and allow only for in-memory content to explicit control
```
akondas, line 12: yup, this class is to magical, while rewriting the tests, I was able to be surprised more than once, because the file system is not defined in the code

tests/Unit/DistTest.php
```diff
@@ -0,0 +1,18 @@
...
+use Buddy\Repman\Service\Dist;
+use PHPUnit\Framework\TestCase;
+
+final class DistTest extends TestCase
+{
+    public function testVersionWithSlash(): void
+    {
+        $dist = new Dist('repo', 'package', 'dev-master/feature', '123456', 'zip');
+
+        self::assertEquals(md5('dev-master/feature'), $dist->version());
```
akondas, line 16: more coverage :muscle:

composer.json
```diff
@@ -71,6 +71,7 @@
         "ekino/phpstan-banned-code": "^0.3.1",
         "friendsofphp/php-cs-fixer": "^2.16",
         "fzaninotto/faker": "^1.9",
+        "league/flysystem-memory": "^1.0",
```
akondas, line 74: this will allow to write cleaner and faster tests

src/Service/Proxy.php
```diff
@@ -110,6 +111,39 @@ public function removeDist(string $package): void
...
+
+    /**
+     * @param mixed[] $files
+     */
+    private function syncPackagesMetadata(array $files): void
+    {
+        foreach ($files as $file) {
+            $url = sprintf('%s://%s', parse_url($this->url, PHP_URL_SCHEME), $file['path']);
+            // todo: what if proxy do not return `Last-Modified` header?
+            if ($this->downloader->getLastModified($url)->getOrElse(0) > ($file['timestamp'] ?? time())) {
```
akondas, line 136: slow performance, maybe we can speed this up with @reactphp :sunglasses:

PR #225, opened 2020-07-17
Implement telemetry settings

karniv00l:


templates/component/telemetryPrompt.html.twig
```diff
@@ -0,0 +1,16 @@
...
+        <form class="card" action="{{ path('admin_config_telemetry_enable') }}" method="POST">
+            <input type="hidden" name="_method" value="POST">
+            <button class="btn btn-info" type="submit">Okay</button>
+        </form>
+        <form class="card" action="{{ path('admin_config_telemetry_disable') }}" method="POST">
+            <input type="hidden" name="_method" value="DELETE">
+            <button class="btn btn-secondary" type="submit">No, thanks</button>
+        </form>
+    </div>
+</div>
```
karniv00l, line 16: Screenshot 2020-07-17 at 18 06 44

src/Controller/Admin/ConfigController.php
```diff
@@ -39,4 +42,30 @@ public function edit(Request $request): Response
...
+     */
+    public function disableTelemetry(Request $request): Response
+    {
+        $this->telemetry->generateInstanceId();
+        $this->dispatchMessage(new ChangeConfig([
+            'telemetry' => 'disable',
+        ]));
+
+        return $this->redirectToRoute('index');
+    }
```
karniv00l, line 70: Screenshot 2020-07-17 at 18 07 44

config/services.yaml
```diff
@@ -85,6 +85,10 @@ services:
             $databaseDir: '%security_advisories_db_dir%'
             $databaseRepo: '%security_advisories_db_repo%'
 
+    Buddy\Repman\Service\Telemetry:
+        arguments:
+            $instanceIdFile: '%kernel.project_dir%/var/instance-id'
```
akondas, line 90: You can bind `$instanceIdFile` variable name to `'%kernel.project_dir%/var/instance-id'`. Then it will be injected automatically :mage_man:

src/Command/CreateAdminCommand.php
```diff
@@ -53,6 +59,17 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             ['ROLE_ADMIN']
         ));
 
+        $question = new ConfirmationQuestion('Allow for sending anonymous usage statistic? (y/n) ', true);
```
akondas, line 62: We will need to add a link to the documentation here, where it says exactly what data is collected.

src/Command/CreateAdminCommand.php
```diff
@@ -53,6 +59,17 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+        $answer = $this
+            ->getHelper('question')
+            ->ask($input, $output, $question);
+
+        $this->bus->dispatch(new ChangeConfig([
+            'telemetry' => $answer ? 'enabled' : 'disabled',
+        ]));
```
akondas, line 69: We need to check if instance id file is already generated. We cannot ask about this option every time.

templates/component/telemetryPrompt.html.twig
akondas: this should only be visible to the instance admin

akondas: ok, now i see :facepalm:

PR #230, opened 2020-07-20
Dockerfile: Install subversion client

pavog:
Install subversion client in dockerfile to enable support for subversion.


akondas:
Thanks :+1:

PR #237, opened 2020-07-24
Implement technical email

karniv00l:


src/Controller/Admin/ConfigController.php
```diff
@@ -32,6 +34,7 @@ public function edit(Request $request): Response
         $form = $this->createForm(ConfigType::class, $this->config->getAll());
         $form->handleRequest($request);
         if ($form->isSubmitted() && $form->isValid()) {
+            $this->handleTechnicalEmail((string) $form->get(Config::TECHNICAL_EMAIL)->getData());
```
akondas, line 37: I'm not sure. In that case it will be possible to change this config and bypass the logic of sending the technical email to the server, it would be better to add it to the handler (`ChangeConfigHandler`).

src/Service/Telemetry/Email.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Buddy\Repman\Service\Telemetry;
+
+final class Email implements \JsonSerializable
```
akondas, line 7: :thinking: maybe `TechnicalEmail`?

PR #238, opened 2020-07-28
Update version date when updating

nickygerritsen:
I noticed that the version thingy I created didn't update the date when a version is updated. This fixes that.


akondas:
Thanks @nickygerritsen :beers:

PR #248, opened 2020-08-20
Update Symfony to 5.1

marmichalski:
Related to #206, #239

PR #250, opened 2020-09-01
Sf5.1 update

karniv00l:
Additions to #248, rebased

### DIFF
--- a/src/Command/CreateAdminCommand.php
+++ b/src/Command/CreateAdminCommand.php
@@ -20,6 +20,8 @@ use Symfony\Component\Messenger\MessageBusInterface;
 
 final class CreateAdminCommand extends Command
 {
+    protected static $defaultName = 'repman:create:admin';
+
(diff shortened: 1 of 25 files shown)


src/Command/CreateAdminCommand.php
```diff
@@ -20,6 +20,8 @@
 
 final class CreateAdminCommand extends Command
 {
+    protected static $defaultName = 'repman:create:admin';
+
```
karniv00l, line 24: This will make command lazy, there was a warning when Telemetry service was injected and database was not created yet (requires messenger.transport.failed)

src/Migrations/Factory/MigrationFactoryDecorator.php
```diff
@@ -0,0 +1,33 @@
+<?php
```
karniv00l, line 1: Allows for using `$this->container` in migrations

akondas (approved):

PR #253, opened 2020-09-02
Add queue for downloader to limit concurrent requests

akondas:


src/Service/Downloader/ReactDownloader.php
```diff
@@ -6,22 +6,32 @@
...
 use React\Http\Browser;
+use React\Promise\PromiseInterface;
 use React\Socket\Connector;
 
 final class ReactDownloader implements Downloader
 {
     private LoopInterface $loop;
     private Browser $browser;
+    private Queue $headQueue;
+    private Queue $getQueue;
```
akondas, line 23: maybe we can handle this in one queue :thinking:

akondas: yup, it is possible :exploding_head:

akondas: Variable method call :disappointed:

karniv00l (approved):

PR #259, opened 2020-09-06
Package search

giggsey:
Just want a bit of feedback/advice before I continue on this:
 - I'm thinking just an initial basic package name search for the time being, could be expanded to description (and maybe tags?) later
 - Currently, the pagination is done as two different calls to the `PackageQuery` class - to avoid having to add search functionality for each, could a paginator bundle be used instead? (I come from a Laravel background where it's a core part of the framework)
 - Design wise, I'm thinking that the search bar could be on the right of the title bar, to the left of 'Add package' 
image
 - Side note: could the hard-coded 20 per page be moved to a user setting?

### DIFF
--- a/src/Controller/OrganizationController.php
+++ b/src/Controller/OrganizationController.php
@@ -72,14 +72,9 @@ final class OrganizationController extends AbstractController
         $filter = Filter::fromRequest($request);
 
         $count = $this->packageQuery->count($organization->id(), $filter);
-        $user = parent::getUser();
 
-        // If the filtered version returned 0 results, we need to count again, but without any filters
-        if ($count === 0 &&
-            $this->packageQuery->count($organization->id(), (new Filter())) === 0 &&
-            $user instanceof User &&
-            $organization->isOwner($user->id())
-        ) {
+        // If the filtered count has no results, we need to check if the organization has no packages
+        if ($count === 0 && $this->hasNoPackages($organization)) {
             return $this->redirectToRoute('organization_package_new', ['organization' => $organization->alias()]);
         }
 
@@ -370,5 +365,14 @@ final class OrganizationController extends AbstractController
                 $this->exceptionHandler->handle($exception);
             }
         }
+    }
+
+    private function hasNoPackages(Organization $organization): bool
+    {
+        $user = parent::getUser();
+
+        return $user instanceof User &&
+            $organization->isOwner($user->id()) &&
+            $this->packageQuery->count($organization->id(), (new Filter())) === 0;
     }
 }
--- a/src/Controller/OrganizationController.php
+++ b/src/Controller/OrganizationController.php
@@ -69,14 +69,22 @@ final class OrganizationController extends AbstractController
      */
     public function packages(Organization $organization, Request $request): Response
     {
-        $count = $this->packageQuery->count($organization->id());
+        $count = $this->packageQuery->count($organization->id()); // @todo
         $user = parent::getUser();
         if ($count === 0 && $user instanceof User && $organization->isOwner($user->id())) {
             return $this->redirectToRoute('organization_package_new', ['organization' => $organization->alias()]);
         }
 
+        $searchTerm = $request->get('search');
+
+        if ($searchTerm) {
+            $packages = $this->packageQuery->find($organization->id(), $searchTerm, 20, (int) $request->get('offset', 0));
+        } else {
+            $packages = $this->packageQuery->findAll($organization->id(), 20, (int) $request->get('offset', 0));
+        }
+
         return $this->render('organization/packages.html.twig', [
-            'packages' => $this->packageQuery->findAll($organization->id(), 20, (int) $request->get('offset', 0)),
+            'packages' => $packages,
             'count' => $count,
             'organization' => $organization,
         ]);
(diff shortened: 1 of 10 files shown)


akondas:
Hey @giggsey, looks like a good idea. 
When it comes to pagination with filters, I usually do it through an additional object. So you can create something like this:

```php
final class Filter
{
    private int $offset;
    private int $limit;

    public function __construct()
    {
        $this->offset = 0;
        $this->limit = 20;
    }

    public function setOffset(int $offset): self
    {
        $this->offset = $offset;

        return $this;
    }

    public function setLimit(int $limit): self
    {
        $this->limit = $limit;

        return $this;
    }

    public function limit(): int
    {
        return $this->limit;
    }

    public function offset(): int
    {
        return $this->offset;
    }
}
```
(you can add `$query` param for serach reason)

And then add it to count and found methods in query:
```php
interface PackageQuery
{
    public function findAll(Filter $filter): array;
    public function count(Filter $filter): int;
}
```

Side note about UI: here is mine proposition :wink: 
Screenshot from 2020-09-07 08-53-46

giggsey:
@akondas Sounds good - I'll continue over the next few days.

giggsey:
image

giggsey:
@akondas Tests are failing because the test data is inserting NULL name & description. I don't use Postgres, and I'm not sure if there is a trick to include NULLs when the Filter searchTerm is empty.

src/Controller/OrganizationController.php
```diff
@@ -69,14 +69,23 @@ public function overview(Organization $organization): Response
...
+
+        $count = $this->packageQuery->count($organization->id(), $filter);
         $user = parent::getUser();
-        if ($count === 0 && $user instanceof User && $organization->isOwner($user->id())) {
+
+        // If the filtered version returned 0 results, we need to count again, but without any filters
+        if ($count === 0 &&
+            $this->packageQuery->count($organization->id(), (new Filter())) === 0 &&
+            $user instanceof User &&
+            $organization->isOwner($user->id())
```
akondas, line 81: maybe we can find a name for this arguments and extract private method, for example `isNoPacakges()` or `suggestAddingPackage()`

src/Query/Filter.php
```diff
@@ -0,0 +1,70 @@
...
+        $offset = (int) $request->get('offset', 0);
+
+        if ($offset < 0) {
+            $offset = 0;
+        }
+
+        return (new self())
+            ->setOffset($offset)
+            ->setLimit($limit)
+            ->setSearchTerm($request->get('search', null));
```
akondas, line 68: What about constructor instead of this setters?
In this way `Filter` can have null offset or limit and that should not be possible. This is probably also the cause of the error in test pipeline.

src/Query/User/PackageQuery/DbalPackageQuery.php
```diff
@@ -49,13 +50,16 @@ public function findAll(string $organizationId, int $limit = 20, int $offset = 0
...
             ORDER BY name ASC
             LIMIT :limit OFFSET :offset', [
                 ':organization_id' => $organizationId,
-                ':limit' => $limit,
-                ':offset' => $offset,
-            ]));
+                ':limit' => $filter->getLimit(),
+                ':offset' => $filter->getOffset(),
+                ':namesearch' => '%'.$filter->getSearchTerm().'%',
+                ':descsearch' => '%'.$filter->getSearchTerm().'%',
```
akondas, line 61: `namesearch` and `descsearch` can be replaced with one variable, there is no need for two identical params.
Also, we need to split this sql string to parts, for example:
```php
        $params = [];
        $sql = 'SELECT
                id,
        // ...
                last_scan_result
            FROM organization_package
            WHERE organization_id = :organization_id';
        
        if($filter->hasSerachTerm()) {
            $sql .= ' AND (name ILIKE :term OR description ILIKE :term)';
            $params[':term'] = $filter->serachTerm();
        }
        
        $sql .= ' GROUP BY id
            ORDER BY name ASC
            LIMIT :limit OFFSET :offset';
```

akondas (commented):
I left some comments, the rest looks fine :clap:

src/Query/Filter.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $this->searchTerm = $searchTerm;
+    }
+
+    public function getOffset(): int
+    {
+        return $this->offset;
+    }
+
+    public function setOffset(int $offset): Filter
```
akondas, line 33: if these setter are not used maybe we will remove them?

akondas (commented):
Nice :+1: 
Do you want to try to write a functional test for this functionality? (you can check `OrganizationControllerTest::testPackages` should be very similar, also in test you can use `$this->fixtures->syncPackageWithData` to populate package with given name)

giggsey:
Sure, I'll look at adding the tests.

src/Query/Filter.php
```diff
@@ -0,0 +1,77 @@
...
+    private int $limit = 20;
+    private ?string $searchTerm;
+
+    public function __construct(int $offset = 0, int $limit = 20, ?string $searchTerm = null)
+    {
+        if ($offset > 0) {
+            $this->offset = $offset;
+        }
+
+        if ($limit >= 0) {
```
akondas, line 21: We should add max limit, to prevent return 1k+ rows in one reqeust. I think `100` should be fine :thinking:

akondas (approved):
Thanks :beers: @giggsey, this will also fix error with negative offset :wink:

giggsey:
@akondas There are other places that use pagination, but are still doing it manually. Want me to convert them over to Filter (no need to add search to them)

akondas:
Whenever you have the opportunity, I would like to. I'm just wondering if I should make a separate filter then so that this search term is not confusing. In the sense, if a particular query does not implement a search, let it use a filter without that capability.

If so, I would suggest to do a separate filter in the namespace of `Package` and leave this general filter only with the limit and offset attributes.

PR #263, opened 2020-09-08
Replace all pagination with Filter classes

giggsey:
Continuation of #259

### DIFF
--- a/src/Controller/OrganizationController.php
+++ b/src/Controller/OrganizationController.php
@@ -24,12 +24,13 @@ use Buddy\Repman\Message\Organization\RemovePackage;
 use Buddy\Repman\Message\Organization\RemoveToken;
 use Buddy\Repman\Message\Organization\SynchronizePackage;
 use Buddy\Repman\Message\Security\ScanPackage;
+use Buddy\Repman\Query\Filter;
 use Buddy\Repman\Query\User\Model\Installs\Day;
 use Buddy\Repman\Query\User\Model\Organization;
 use Buddy\Repman\Query\User\Model\Package;
 use Buddy\Repman\Query\User\OrganizationQuery;
 use Buddy\Repman\Query\User\PackageQuery;
-use Buddy\Repman\Query\User\PackageQuery\Filter;
+use Buddy\Repman\Query\User\PackageQuery\Filter as PackageFilter;
 use Buddy\Repman\Security\Model\User;
 use Buddy\Repman\Service\ExceptionHandler;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
@@ -69,7 +70,7 @@ final class OrganizationController extends AbstractController
      */
     public function packages(Organization $organization, Request $request): Response
     {
-        $filter = Filter::fromRequest($request);
+        $filter = PackageFilter::fromRequest($request);
 
         $count = $this->packageQuery->count($organization->id(), $filter);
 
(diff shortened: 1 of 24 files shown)


giggsey:
I force pushed to try to get rid of some of the unrelated commits, but I think the pipeline is from before. CS Fixer runs fine locally

src/Controller/OrganizationController.php
```diff
@@ -24,12 +24,12 @@
 use Buddy\Repman\Message\Organization\RemoveToken;
 use Buddy\Repman\Message\Organization\SynchronizePackage;
 use Buddy\Repman\Message\Security\ScanPackage;
-use Buddy\Repman\Query\Filter;
 use Buddy\Repman\Query\User\Model\Installs\Day;
 use Buddy\Repman\Query\User\Model\Organization;
 use Buddy\Repman\Query\User\Model\Package;
 use Buddy\Repman\Query\User\OrganizationQuery;
 use Buddy\Repman\Query\User\PackageQuery;
+use Buddy\Repman\Query\User\PackageQuery\Filter;
```
akondas, line 32:
```suggestion
use Buddy\Repman\Query\User\PackageQuery\Filter as PackageFilter;
```

src/Query/User/PackageQuery.php
```diff
@@ -4,13 +4,13 @@
 
 namespace Buddy\Repman\Query\User;
 
-use Buddy\Repman\Query\Filter;
```
akondas, line 7: same as above, use alias :wink:

src/Query/User/PackageQuery.php
```diff
@@ -22,6 +22,11 @@ public function findAll(string $organizationId, Filter $filter): array;
 
     public function count(string $organizationId, Filter $filter): int;
 
+    /**
+     * @return Package[]
+     */
+    public function find(string $organizationId, string $searchString, int $limit = 20, int $offset = 0): array;
```
akondas, line 28: :thinking: looks like this isn't used anywhere

src/Query/User/PackageQuery/Filter.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public function hasSearchTerm(): bool
+    {
+        return $this->searchTerm !== null;
+    }
+
+    public function getQueryStringParams(): array
+    {
+        $params = \Buddy\Repman\Query\Filter::getQueryStringParams();
```
akondas, line 35: please, try always to avoid FQCN

src/Query/User/PackageQuery/Filter.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @author giggsey
+ */
```
akondas, line 6: I suspect that it is inserted from template :wink:

akondas (commented):
minor touches to make, I improved the code style but now phpstan is still complaining about something, is the status of this build visible to you?

giggsey:
Both PHPStan failures are because `TokenFixtures` and `PrivatePackageDownloadFixtures` want to load 9999 organisations. From what I can see, both classes are only used to populate dummy data. Happy to them to limit to 100 instead?

akondas (approved):
Thanks :beers: :cocktail:

src/Query/User/PackageQuery/DbalPackageQuery.php
```diff
@@ -144,7 +143,7 @@ public function versionCount(string $packageId): int
     /**
      * @return Version[]
      */
-    public function getVersions(string $packageId, int $limit = 20, int $offset = 0): array
+    public function getVersions(string $packageId, \Buddy\Repman\Query\Filter $filter): array
```
akondas, line 146: I wonder if we can somehow automate the detection of this notation?

giggsey: Possibly 'fully_qualified_strict_types', but it looks like that only works when the use statement is there in the first place

akondas:
Thanks @giggsey for your work :beers:

PR #264, opened 2020-09-09
Update doctrine-bundle and remove deprecation notice

akondas:
This will remove deprecation notice:
```
Since symfony/messenger 5.1: The "Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory" class is deprecated, use "Symfony\Component\Messenger\Bridge\Doctrine\Transport\DoctrineTransportFactory" instead. The Doctrine transport has been moved to package "symfony/doctrine-messenger" and will not be included by default in 6.0. Run "composer require symfony/doctrine-messenger".
```
Screenshot from 2020-09-09 10-33-22

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
     "name": "repman-io/repman",
     "type": "project",
-    "description": "PHP Repository Manager - fast packagist proxy",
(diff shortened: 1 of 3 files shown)


akondas:
Closing, since there is a problem with this ... :disappointed:

PR #283, opened 2020-09-23
Set packages.json last modified based on a provider-latest timestamp

karniv00l:


src/Controller/ProxyController.php
```diff
@@ -54,12 +54,15 @@ public function packages(Request $request): JsonResponse
                 ],
             ],
             'providers-lazy-url' => '/p/%package%',
-            'provider-includes' => $providerHash !== null ? ['p/provider-latest$%hash%.json' => ['sha256' => $providerHash]] : [],
+            'provider-includes' => $metadata->isPresent() === true ? ['p/provider-latest$%hash%.json' => ['sha256' => $metadata->get()->hash()]] : [],
```
akondas, line 57: `$metadata->isPresent() === true` why not just `$metadata->isPresent()`

akondas (approved):

Issue #287, opened 2020-09-23
Limit number of releases?


temp:
We just added a large repository from our self-hosted GitLab, it has about 1500 tags (yes, it's a rather old project...).
repman wants to create a zip for every tag, which will end in 1500 zips with each 50 mb, resulting in 75 gb of release-zips, of which the first 1499 will never be used ;-)
Is there a way to restrict the number of historical releases that are handled?

akondas:
Downloading all distributions ahead of time was purposeful (@karniv00l maybe you remember why?). In fact, this use case complicates things a bit. We could make a separate option in the instance settings to help with this:

`Download dist on sync` - if yes then all distibution will be downloaded during synchronization, if no, only metadata (version history) will be downloaded. Default option will be `no`.

Let me know what you think.

temp:
Would work for me!
What I did on our instance was delete all old zips by hand... I know, not an ideal solution, but at least the storage didn't explode :-D

temp:
But this should be configurable during package-add...

akondas:
Hey @temp, we are almost at the end:
![Screenshot from 2020-10-08 08-40-00](https://user-images.githubusercontent.com/8239917/95423759-ebe35300-0941-11eb-83f8-7d288d75acd3.png)

@karniv00l one more thing:
- [ ] allow to update this limit

giggsey:
@karniv00l Should #294 limit what's shown on the UI or the latest release?

I think the serialised metadata itself is fine, just the DB stored data.

Syncing 3 releases of https://github.com/symfony/finder.git provides:
![image](https://user-images.githubusercontent.com/305730/95866389-c39f8e00-0d5f-11eb-917d-21ab9758c512.png)
![image](https://user-images.githubusercontent.com/305730/95866603-06616600-0d60-11eb-9e20-fe9f994552cf.png)

(Top screenshot app.repman.io, bottom local). Both added with same URL + package restrictions today.

karniv00l:
@giggsey we are limiting number of versions and distributions files, this includes dev and stable releases.I will look into this, thanks

temp:
Just upgrade to 1.1.0, and will try the feature now. What happens when I limit the releases afterwards? Will it throw away the obsolete releases?

karniv00l:
@temp It will remove old releases

temp:
Awesome, thanks! Works fine for me!
... Job just finished, all old releases are gone, and storage was freed. Great work!

PR #294, opened 2020-09-30
Add option to limit number of package versions being imported

karniv00l:
As requested in #287

### DIFF
--- a/src/Entity/Organization/Package.php
+++ b/src/Entity/Organization/Package.php
@@ -112,18 +112,18 @@ class Package
     /**
      * @ORM\Column(type="integer")
      */
-    private int $versionsLimit = 0;
+    private int $keepLastReleases = 0;
 
     /**
      * @param mixed[] $metadata
      */
-    public function __construct(UuidInterface $id, string $type, string $url, array $metadata = [], int $versionsLimit = 0)
+    public function __construct(UuidInterface $id, string $type, string $url, array $metadata = [], int $keepLastReleases = 0)
     {
         $this->id = $id;
         $this->type = $type;
         $this->repositoryUrl = $url;
         $this->metadata = $metadata;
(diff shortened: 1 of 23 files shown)


src/Form/Type/Organization/AddPackageType.php
```diff
@@ -62,6 +63,12 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                     new NotNull(),
                 ],
             ])
+            ->add('versionsLimit', IntegerType::class, [
+                'label' => 'Limit number of versions',
+                'data' => 0,
+                'help' => 'Put "0" to download all versions',
```
akondas, line 69: we need more description here, what limit means to end user?

src/Entity/Organization/Package.php
```diff
@@ -109,15 +109,21 @@ class Package
      */
     private Collection $versions;
 
+    /**
+     * @ORM\Column(type="integer")
+     */
+    private int $versionsLimit = 0;
```
akondas, line 115: :thinking: or maybe `keepLastReleases`?

src/Form/Type/Organization/AddPackageType.php
```diff
@@ -62,6 +63,12 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                     new NotNull(),
                 ],
             ])
+            ->add('versionsLimit', IntegerType::class, [
+                'label' => 'Limit number of versions',
```
akondas, line 67: technically, this is not limit, this is number of releases that will be downloaded from source ,and we download them to speed up install process. So maybe `Download latest releases` :thinking:

akondas (approved):

akondas (approved):
LGTM :+1:

PR #298, opened 2020-10-13
Improve Package Details UX

giggsey:
image

image

image


giggsey:
I think the header should get a bit more consistent across these pages too.

"[Package name] [page]" with Package name being a link back to the Details.

Not sure about Package list being a button on each of those pages

akondas (approved):

akondas:
Tested :+1:, UX feels much better, thanks @giggsey.
I like your idea with small package button on the left. You can change this as well :wink:

giggsey:
image

image

Using JavaScript to format the numbers to allow for different locales without having to build support into the backend

akondas:
Nice, link to stats in package details is also good UX :+1:

PR #300, opened 2020-10-14
Allow package list to be sortable

giggsey:
image

### DIFF
--- a/src/Query/Filter.php
+++ b/src/Query/Filter.php
@@ -10,8 +10,10 @@ class Filter
 {
     private int $offset = 0;
     private int $limit = 20;
+    private ?string $sortColumn = null;
+    private string $sortOrder;
 
-    public function __construct(int $offset = 0, int $limit = 20)
+    public function __construct(int $offset = 0, int $limit = 20, ?string $sort = null)
     {
         if ($offset >= 0) {
             $this->offset = $offset;
@@ -24,6 +26,15 @@ class Filter
         if ($this->limit > 100) {
             $this->limit = 100;
         }
+
+        // $sort = 'column:order'
+        $sortParts = explode(':', $sort ?? '');
+        if ($sortParts[0] !== '') {
+            $this->sortColumn = $sortParts[0];
+        }
+
+        $sortOrder = $sortParts[1] ?? 'asc';
+        $this->sortOrder = in_array($sortOrder, ['asc', 'desc'], true) ? $sortOrder : 'asc';
     }
 
     public function getOffset(): int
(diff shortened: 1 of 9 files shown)


giggsey:
Not sure where the existing icons came from, I believe https://github.com/lucide-icons/lucide

tests/Functional/Controller/OrganizationControllerTest.php
```diff
@@ -191,6 +191,67 @@ public function testPagination(): void
...
+
+        // Sort by invalid column
+        $this->client->request('GET', $this->urlTo('organization_packages', ['organization' => 'buddy', 'limit' => 1, 'sort' => 'invalid-column:asc']));
+
+        self::assertTrue($this->client->getResponse()->isOk());
+        $content = (string) $this->client->getResponse()->getContent();
+        self::assertStringContainsString('buddy-works/package-1', $content);
+        self::assertStringContainsString('sort=name:asc', $content);
+        self::assertStringContainsString('sort=version:asc', $content);
+        self::assertStringContainsString('sort=date:asc', $content);
```
akondas, line 252: nice coverage :+1:

akondas (approved):
LGTM, @karniv00l check this please and merge if looks good to you :wink:

src/Query/Filter.php
```diff
@@ -36,22 +47,44 @@ public function getLimit(): int
...
+
+        if ($this->hasSort()) {
+            $return['sort'] = $this->sortColumn.':'.$this->sortOrder;
+        }
+
+        return $return;
     }
 
-    public static function fromRequest(Request $request): self
+    public static function fromRequest(Request $request, ?string $defaultSortColumn = null): self
```
akondas, line 82: just thinking, maybe we can put `name` here as default?

giggsey: At the moment only the Organisation Package list supports sorting, but other tables could. I think it's clearer if the Controller specified the default, to avoid any confusion with the database layer getting a default that it doesn't actually handle

akondas: :+1: solid argument

karniv00l:
> Not sure where the existing icons came from, I believe https://github.com/lucide-icons/lucide

They are from https://feathericons.com/ but with `class="icon"` attribute

giggsey:
> > Not sure where the existing icons came from, I believe https://github.com/lucide-icons/lucide
> 
> They are from https://feathericons.com/ but with `class="icon"` attribute

Cool, Lucide is a community fork of Feature Icons.

Happy to keep those icons, and do they need attribution anywhere?

karniv00l (approved):

karniv00l:
> > > Not sure where the existing icons came from, I believe https://github.com/lucide-icons/lucide
> > 
> > 
> > They are from https://feathericons.com/ but with `class="icon"` attribute
> 
> Cool, Lucide is a community fork of Feature Icons.
> 
> Happy to keep those icons, and do they need attribution anywhere?

It doesn't matter to me which icons we use, your call, but I think that we need to include licences

giggsey:
Feather is MIT, and Lucide is ISC (very similar to MIT). Both licenses require the license text to be included with the SVG distribution. As the svgs are embedded and distributed on every request, my understanding is that the license needs to be included there too.

Not sure how to continue here. Maybe a separate page listing third-party software & licenses.

akondas:
Thanks @giggsey, we will merge it, and @karniv00l will add required licenses text to readme.

PR #303, opened 2020-10-15
Display README.md for packages

giggsey:
Extract the file from the latest release dist zip, and store rendered in the database.

image

Part of #175


src/Query/User/PackageQuery/DbalPackageQuery.php
```diff
@@ -143,7 +143,8 @@ public function getById(string $id): Option
                 last_scan_date,
                 last_scan_status,
                 last_scan_result,
-                keep_last_releases
+                keep_last_releases,
+                readme
```
akondas, line 147: :thinking: since it will be displayed in one place anyway, we could do a separate read model for it ...
one could even slim down the current model `Package` and create `PackageDetails` with readme, install and other this that are required **only** on package details page.

giggsey: getByID() is only used in places where the package ID is a route parameter. These is all the frontend package 'action' pages, and the webhook endpoint.

Out of those routes, the package details is going to be the most often used. If PackageDetails was split out, then this page would be querying the same table twice.

Unless I create a new route model binding (I come from Laravel, so not sure of the Symfony naming) for PackageDetails, and have that model extend Package.

giggsey:
I'll work on the tests, then it'll be ready for review.

src/Query/User/PackageQuery/DbalPackageQuery.php
akondas: You can create `PackageDetailsParamConverter` for that (look at `PackageParamConverter`)

giggsey:
The number of release restrictions is affecting how this works.

At the moment, the latest version won't necessarily be in the list of available versions. I think that it should *always* be in that list.

src/Service/Markdown.php
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Buddy\Repman\Service;
+
+use League\CommonMark\CommonMarkConverter;
+use League\CommonMark\Environment;
+use League\CommonMark\Extension\ExternalLink\ExternalLinkExtension;
+use League\CommonMark\Extension\HeadingPermalink\HeadingPermalinkExtension;
+use League\CommonMark\MarkdownConverterInterface;
+
+final class Markdown
```
akondas, line 13: `MarkdownToHtmlConverter` maye? although it is only used in one place

src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
```diff
@@ -207,4 +225,28 @@ private function createConfig(Package $package, IOInterface $io): Config
 
         return $config;
     }
+
+    private function loadREADME(Dist $dist): ?string
```
akondas, line 229: I think this should be moved to a separate class, or to `Markdown` service. Maybe then you can name it `ReadmeExtractor` or something like that.

src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
```diff
@@ -50,6 +54,15 @@ public function synchronize(Package $package): void
             $json = ['packages' => []];
             $packages = $repository->getPackages();
 
+            // @todo Could we use \Composer\Semver\Semver::sort() instead?
+            usort($packages, static function (PackageInterface $a, PackageInterface $b): int {
```
akondas, line 58: I need to talk it over with @karniv00l, but initially yes :wink:

tests/Unit/Service/PackageSynchronizer/ReadmeComposerPackageSynchronizerTest.php
```diff
@@ -0,0 +1,125 @@
...
+        self::assertCount(1, $package->versions());
+        self::assertEquals('1.3.0', $package->latestReleasedVersion());
+        self::assertEquals("<h1><a id=\"user-content-test\" href=\"#test\" name=\"test\" class=\"\" aria-hidden=\"true\" title=\"\"></a>Test</h1>\n<p>Testing</p>\n", $package->readme());
+    }
+
+    /**
+     * Artifacts imported should have the files in the top level directory
+     * Copies from VCS have a top level directory of the project name & commit ID
+     * This will test loading the files from that second directory, but will use artifact
+     * to make the test simpler.
```
akondas, line 94: thanks for the clarification :+1:

akondas (approved):

src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
giggsey: Done.

As it's now its own class, it could be tested separately from the ComposerPackageSyncronizer. I think for now, we could leave the existing tests I added, as they are testing selecting the correct version.

If further readme files are added in the future, then they should be in a dedicated unit test.

src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
karniv00l: @giggsey What's your intention here? You already did more than `sort()` provides

giggsey:
The usort fixes a bug with the initial $latest being ahead of a valid
stable release.

The Composer sort function deals with Dev releases, which becomes more
important when you start restricting the number of releases.

src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
giggsey: Reply via email didn't work...

The usort fixes a bug with the initial $latest being ahead of a valid stable release.

The Composer sort function deals with Dev releases, which becomes more important when you start restricting the number of releases.

giggsey: I just experimented with this and the only difference is that it normalises the versions first. But the versions are already normalised in the $packages list.

I'll remove the todo.

karniv00l:
Thanks @giggsey, will you resolve the conflict with the master also?

giggsey:
Ignore that last commit, I think the merge went bad.

I'll also downgrade my composer back to v1 to do this

giggsey:
That `composer update` did not go well...

ad4b80a was actually fine, but I've done that update again using Composer v1.

Should be ready now @karniv00l

karniv00l:
Good job @giggsey 🍺

giggsey:
Thanks for merging.

I go back to work tomorrow after paternity leave, so I should be leaving you guys in peace for a bit ;)

PR #305, opened 2020-10-16
Update doctrine-bundle and symfony to remove deprecation notice

akondas:
This will fix this deprecation notice:
Screenshot from 2020-10-16 21-40-19

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "clue/mq-react": "^1.2",
         "composer/composer": "^1.10.1",
         "composer/semver": "^1.5",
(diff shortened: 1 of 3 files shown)


akondas:
Pipeline failed:
```
                                                                               
In PDOConnection.php line 41:
                                                                               
  SQLSTATE[25P01]: No active sql transaction: 7 ERROR:  LOCK TABLE can only b  
  e used in transaction blocks     
```

This is actually quite funny because I fixed the bug myself :wink:

karniv00l (approved):

PR #315, opened 2020-10-20
Fix artifact repo security scan

giggsey:
Non VCS repos don't have a reference. Repman uses the shasum instead for the Dist filename.


giggsey:
image

akondas (approved):

akondas:
Thanks @giggsey :beers:

PR #366, opened 2021-01-07
Remove Storage interface and replace by StorageImpl

pedro-stanaka:
This is the cleanup I promised, removing the Storage interface to avoid confusion since it will only have one implementation. Moreover, I didnt make the class final, otherwise we won't be able to mock it with prophecy lib.

Feel free to close/ignore if you think this is not needed.


akondas (commented):
I'm just wondering about the namespace. I really like the changes themselves. What about:
```
Buddy\Repman\Service\Dist\Storage
```
(without second `Storage`)

I apologize in advance for the late pace, but this week will still be quite busy for me :wink:

pedro-stanaka:
No worries, we all have those weeks. I moved the class now, the namespace was strange really.

akondas (approved):

akondas:
Thanks @pedro-stanaka :clap:

PR #373, opened 2021-01-08
Introducing cached adapters to reduce IO/HTTP overhead with file handling

### LINKED ISSUE #359: Improve performance of file operations by introducing cached adapters for Flysystem
After #332 gets merged we could think of an improvement of performance by adding a [cached adapter for flysystem](https://flysystem.thephpleague.com/v1/docs/advanced/caching/).

When using S3, this reduces the delay on checking whether some file is present or not on S3 and avoid HTTP chatter.

pedro-stanaka:
Closes #359. Small improvement to cache the stats of files being handled inside the application to avoid unnecessary IO or, in worst case, HTTP chatter.


akondas:
@karniv00l please check, if everything is ok then there is a green light on my part :wink:

karniv00l (approved):

akondas (approved):

PR #376, opened 2021-01-18
Disable by default cached adapters

akondas:


akondas:
Temporary fix until we find no workaround for the memory limit issue.

Issue #381, opened 2021-01-18
Onboarding new user/organization


noniagriconomie:
Hi 👋 

First of all thank you for this awesome projet, we just began to use it at work 👍 

---

I just wanted to share with you the onboarding process that was a little bit funky 😄 

Process to reproduce:

- a collegue of mine created an account + organization
- he invites me, I received an /invitation link
- the invitation link redirects to /login, first why not /register?
- then after clicking /register (via github), I did not saw that I was already part of an organization and the GUI invits me to create an organization, I had to close this page and open the https://app.repman.io to see that I am in the org actually

Obviously, I can propose/make some changes if you found this relevant 😃 

Many thanks,

akondas:
Sure, we will gladly accept the improvement of this flow. What do you propose as a solution?

noniagriconomie:
@akondas Hi, thanks for your reply

2 things as the `/user/invitation` is protected by the firewall right now:

- either inform the user clicking its invitation link to first register or login then to accept the invitation

this could be changed by rephrasing the email content for example

```
XXX has invited you to use Repman to collaborate with them. Create an account first or login and then use the button below to accept invitation and get started with XXX organization.
```

- or perhaps this url could be allowed as non logged in like the reset password url
https://github.com/repman-io/repman/blob/9da2e58375dd9417da558401d7a5f2dabf71cd42/src/Controller/Organization/MembersController.php#L55 

The user receiving this invitation link and clicking it (and after doing the actual job) can be then redirected to the organization (default) or login (with a flash message telling him the invitation was successful) and should login with the same email for example

akondas:
I think that the best solution would be to add an invitation token to the session and check if this token exists after successful logging in / registration and if so, redirect the user to the path `organization_accept_invitation`. What do you think about it?

noniagriconomie:
@akondas sound great yes, I will be glad to try this soon and propose a PR

PR #387, opened 2021-01-22
Improve organization invitation with registration/login flow

noniagriconomie:
Hello

Here is a PR that tries to improve the organization invitation process flow
fixes https://github.com/repman-io/repman/issues/381

Friendly ping @akondas 

Thank you,

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -54,6 +54,7 @@ security:
 
     access_control:
          - { path: ^/admin, roles: ROLE_ADMIN }
+         - { path: ^/user/invitation, roles: [ROLE_USER, IS_AUTHENTICATED_ANONYMOUSLY] }
          - { path: ^/user, roles: ROLE_USER }
          - { path: ^/$, roles: ROLE_USER }
          - { path: ^/organization/.+/overview$, roles: ROLE_ORGANIZATION_ANONYMOUS_USER }
--- a/src/Controller/Organization/MembersController.php
+++ b/src/Controller/Organization/MembersController.php
@@ -21,16 +21,19 @@ use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
 final class MembersController extends AbstractController
 {
+    private SessionInterface $session;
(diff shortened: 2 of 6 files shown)


akondas (commented):
Although I like this change, what do you say to leave one button and do it like this:

The route "organization_accept_invitation" could be opened (available without logging in). In its action, it could be detected if the user is logged in, if not a redirection with 'invitation` token to login (and then the same on the login action, you would have to render the button for registration with `invitation` token)

noniagriconomie:
@akondas thanks for your review
agreed, I will refactor like this, will try to do it this week

noniagriconomie:
@akondas code is rewritten, I am testing it now
new round of review appreciated

src/Controller/Organization/MembersController.php
```diff
@@ -52,12 +55,20 @@ public function listMembers(Organization $organization, Request $request): Respo
     }
 
     /**
-     * @Route("/user/invitation/{token}", name="organization_accept_invitation", methods={"GET"}, requirements={"token"="%uuid_pattern%"})
```
akondas, line 55: instead of changing the path (b.c. break), I would prefer to add an entry to `security.yaml`:
```
- { path: ^/user/invitation., roles: [ROLE_USER, IS_AUTHENTICATED_ANONYMOUSLY] }
```

akondas (commented):
One change and we are ready to go :wink: :+1:

src/Controller/Organization/MembersController.php
noniagriconomie: absolutly right, i have reverted and added the security access control entry

akondas (approved):
:+1: LGTM

config/packages/security.yaml
```diff
@@ -54,6 +54,7 @@ security:
 
     access_control:
          - { path: ^/admin, roles: ROLE_ADMIN }
+         - { path: ^/user/invitation., roles: [ROLE_USER, IS_AUTHENTICATED_ANONYMOUSLY] }
```
akondas, line 57: this `.` is for a purpose?

akondas:
@noniagriconomie Also let me know if you will be able to do a rebase or can I help with it?

config/packages/security.yaml
noniagriconomie: I've just cp your suggestion from https://github.com/repman-io/repman/pull/387#discussion_r566313300
Should I remove it?
Also PR rebased on master

akondas: Ok, it's my mistake then :wink:, please remove it, and we can merge this :+1:

akondas: and sorry for the trouble, I wrote it from memory

noniagriconomie:
@akondas PR rebased and comment addressed for the security access control

akondas (approved):

akondas:
Thanks @noniagriconomie :pray:

Issue #393, opened 2021-01-28
Remove RefreshOAuthTokenMiddleware


akondas:
Long story short: `RefreshOAuthTokenMiddleware` must be removed due to the inability to return the error when there is a problem with refreshing the token. Things to do later:
- [ ] Add abstraction for `ClientRegistry`
- [x] Refactor some controller actions to reduce duplication
- [ ] Refactor `ComposerPackageSynchronizer` :thinking: (hard, but this service grows too much)
- [x] Add info about webhooks to Package (new field when we can put errors)

PR #395, opened 2021-01-29
Refresh oauth token in runtime without failing message or redirect

akondas:
Details in #393
Further work required :wink:


src/Controller/RepoController.php
```diff
@@ -85,7 +85,7 @@ public function distribution(Organization $organization, string $package, string
         return new StreamedResponse(function () use ($filename): void {
             $outputStream = \fopen('php://output', 'wb');
             if (false === $outputStream) {
-                throw new HttpException(500, 'Could not open output stream to send binary file.');
+                throw new HttpException(500, 'Could not open output stream to send binary file.'); // @codeCoverageIgnore
```
akondas, line 88: I have no idea how this could be tested :man_shrugging:

src/Entity/User/OAuthToken.php
```diff
@@ -87,32 +88,18 @@ public function isType(string $type): bool
         return $this->type() === $type;
     }
 
-    public function accessToken(): string
+    public function accessToken(ClientRegistry $oauth): string
```
akondas, line 91: this will be replaced in future with custom service to avoid coupling with `OAuth2ClientBundle`

karniv00l (approved):

PR #414, opened 2021-02-12
Test compatibility issues and small deprecation fix

pedro-stanaka:
Hey guys, I stumbled upon a couple of problems while running the tests on docker (with alpine images), so I wanted to make this little change to make tests behavior better on newer versions of git and different linux distros.

Also, the factory method for Config from phpcs has been deprecated, I took time to fix that as well.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,8 +1,8 @@
 <?php
 
 $finder = PhpCsFixer\Finder::create()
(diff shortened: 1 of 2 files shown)


karniv00l (approved):

akondas (approved):
Thanks :+1:

PR #422, opened 2021-03-01
Fix: composer 9999999-dev issue

### LINKED ISSUE #421: [BUG] 404 error when package version is dev-master
### Configuration
Repman => 1.2.1
Private gitlab => 13.2.2 
php => 7.4
composer => 2.0.9

### The bug
We have a private package hosted in a self-host Gitlab (myvendor/mypackage for example)
When we run : composer require myvendor/mypackage:dev-master
Composer tries to retrieve the dist package from Repman via this URL :
https://organization.repo.repman.domain.com/dists/myvendor/mypackage/dev-master/XXXXXX.zip (where XXXXXX is the commit hash)

Repman answers with a 404 error. After some investigations, we found that Repman is looking for a zip file with this path :
/var/www/repman/shared/var/repo/organization/dist/myvendor/mypackage/dev-master_XXXXXX.zip
but in reality only this file exists :
/var/www/repman/shared/var/repo/organization/dist/myvendor/mypackage/9999999-dev_XXXXXX.zip

It seems related to a Composer issue :
[[2.0][RFC] Stop normalizing dev-master as 9999999-dev](https://github.com/composer/composer/issues/8323)

We add in Service/Dist.php => version() a condition to translate dev-master to 999999-dev.

Is this condition the better solution or it needs to be corrected in an another function ?

slappyslap:
Closes #421

### DIFF
--- a/src/Service/Dist.php
+++ b/src/Service/Dist.php
@@ -37,6 +37,10 @@ final class Dist
             return md5($this->version);
         }
 
+        if ($this->version === 'dev-master') {
+            return '9999999-dev';
+        }
+
         return $this->version;
     }
 
--- a/tests/Unit/DistTest.php
+++ b/tests/Unit/DistTest.php
@@ -15,4 +15,11 @@ final class DistTest extends TestCase
 
         self::assertEquals(md5('dev-master/feature'), $dist->version());
     }
+
+    public function testVersionDevMaster(): void
+    {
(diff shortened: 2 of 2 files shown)


akondas:
Thanks for PR, we will check if it does not cause any BC breaks and let you know.

akondas (approved):
Tested with v1 and v2, everything seems to be working fine.
Thanks for contribution :+1:

PR #425, opened 2021-03-08
Improve PHPUnit assertions

peter279k:
# Changed log

- Using the `assertFileExists` to assert expected file path is existed.
- Using the `assertFileNotExists` to assert expected file path is not existed.
- Using the `assertIsResource` to assert expected type is `resource`.
- Using the `assertCount` to assert expected counter is same as result.

### DIFF
--- a/tests/Functional/Command/ProxySyncReleasesCommandTest.php
+++ b/tests/Functional/Command/ProxySyncReleasesCommandTest.php
@@ -35,7 +35,7 @@ final class ProxySyncReleasesCommandTest extends FunctionalTestCase
         $commandTester = new CommandTester($command);
         $result = $commandTester->execute([]);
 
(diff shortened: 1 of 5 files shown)


akondas:
Thanks for contribution, `assertFileNotExists` is deprecated, `assertFileDoesNotExist` should be used instead: https://github.com/sebastianbergmann/phpunit/issues/4077

peter279k:
> Thanks for contribution, `assertFileNotExists` is deprecated, `assertFileDoesNotExist` should be used instead: [sebastianbergmann/phpunit#4077](https://github.com/sebastianbergmann/phpunit/issues/4077)

You're right. I will fix that.

akondas (approved):

akondas:
Thanks :beers:

PR #477, opened 2021-06-01
Make domain separators configurable

JeckLabs:
First of all, thanks for your amazing work.
Right now, I'm working on a helm chart to deploy the Repman in the Kubernetes cluster.
Among other things, I want to configure domain separators via environmental variables. 
So this is what this pull request for.

### DIFF
--- a/.env
+++ b/.env
@@ -15,6 +15,8 @@
 
 ###> app ###
 APP_HOST=repman.wip
+APP_DOMAIN_SEPARATOR=.
+APP_ORGANIZATION_SEPARATOR=.
 APP_URL_SCHEME=https
 APP_ALLOWED_PACKAGE_TYPES=git,github,gitlab,bitbucket,subversion,mercurial,pear,artifact,path
 APP_GITLAB_API_URL=https://gitlab.com
--- a/.env.docker
+++ b/.env.docker
@@ -19,6 +19,8 @@ PHP_URL=app:9000
 
 ###> app ###
 APP_HOST=repman.wip
+APP_DOMAIN_SEPARATOR=.
+APP_ORGANIZATION_SEPARATOR=.
 APP_URL_SCHEME=https
 APP_ALLOWED_PACKAGE_TYPES=git,github,gitlab,bitbucket,subversion,mercurial,pear,artifact,path
 APP_GITLAB_API_URL=https://gitlab.com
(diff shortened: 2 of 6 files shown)


akondas (changes requested):
Hi, thanks for PR. Looks ok, but you need also add changes in:
- [x] config/packages/security.yaml (firewall and access control)

JeckLabs:
Sorry, I missed that. It should be fixed now. Thanks!

akondas (approved):
LGTM :+1:

karniv00l (approved):

akondas:
One more, we need to make sure that this change is backward compatible. Please add to:
```
organization_separator: '%env(resolve:APP_ORGANIZATION_SEPARATOR)%'
```
default fallback for `domain_separator`. Maybe `APP_ORGANIZATION_SEPARATOR` should be commented out by default?


CC @jmalinens

jmalinens:
> One more, we need to make sure that this change is backward compatible. Please add to:
> 
> ```
> organization_separator: '%env(resolve:APP_ORGANIZATION_SEPARATOR)%'
> ```
> 
> default fallback for `domain_separator`. Maybe `APP_ORGANIZATION_SEPARATOR` should be commented out by default?
> 
> CC @jmalinens

Maybe something like that should be enough but I don't see it as huge BC break and it can be left "as is"

```yaml
parameters:
  env(APP_ORGANIZATION_SEPARATOR): '%domain_separator%'
```

(if APP_ORGANIZATION_SEPARATOR env variable is not set up it will fall back to domain_separator param)

akondas:
@jmalinens Ok, if this does not apply to you, I can accept it even as it is now.

JeckLabs:
@jmalinens, I see possible issues for docker users. They use separate `.env.docker` and would not have a default value for `APP_ORGANIZATION_SEPARATOR` after an upgrade. Probably I should add smth like `%env(default:domain_separator:resolve:APP_ORGANIZATION_SEPARATOR)` in config — will test it later tonight.

JeckLabs:
@akondas, check the latest commit, pls. It should fix compatibility issues. I checked in the docker environment.

akondas (approved):

akondas:
Thanks @JeckLabs :+1:

akondas:
I will merge tomorrow (if that's not a problem), today i have a day off :wink:

PR #485, opened 2021-06-10
Add package url

noniagriconomie:
Add package url small shortcut to package detail page
HTML elements taken from the `templates/organization/packages.html.twig` view

### DIFF
--- a/src/Query/User/Model/PackageDetails.php
+++ b/src/Query/User/Model/PackageDetails.php
@@ -10,6 +10,7 @@ final class PackageDetails
 
     private string $id;
     private string $organizationId;
+    private string $url;
     private ?string $name;
     private ?string $description;
     private ?string $latestReleasedVersion;
@@ -21,6 +22,7 @@ final class PackageDetails
     public function __construct(
         string $id,
         string $organizationId,
+        string $url,
         ?string $name = null,
         ?string $latestReleasedVersion = null,
         ?\DateTimeImmutable $latestReleaseDate = null,
@@ -32,6 +34,7 @@ final class PackageDetails
     ) {
         $this->id = $id;
         $this->organizationId = $organizationId;
+        $this->url = $url;
         $this->name = $name;
         $this->latestReleasedVersion = $latestReleasedVersion;
         $this->latestReleaseDate = $latestReleaseDate;
@@ -50,6 +53,11 @@ final class PackageDetails
     public function organizationId(): string
     {
         return $this->organizationId;
+    }
+
+    public function url(): string
+    {
+        return $this->url;
     }
 
     public function name(): ?string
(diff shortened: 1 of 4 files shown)


akondas:
You need to add `url` to `PackageDetails` read model. Generally, I was looking for it recently, so if the build is ok, we can accept it.

noniagriconomie:
@akondas Sorry I've not described that I not yet tested it locally, just opening the PR for now via copy/past
I will finish it very soon, thank you

src/Query/User/Model/PackageDetails.php
```diff
@@ -11,6 +11,7 @@ final class PackageDetails
     private string $id;
     private string $organizationId;
     private ?string $name;
+    private ?string $url;
```
noniagriconomie, line 14: I've made it nullable as the `name` but seems it can not be null in the model
should I change it?
also what about making `name` not nullable with a fallback to the url if not present when flushing?

akondas: Package url can't be null (when you add package to repman, you must provide url). The name is another matter. It is extracted on sync, which may fail and therefore may be null.

templates/organization/package/details.html.twig
```diff
@@ -20,6 +20,13 @@
             <div class="col-6">
                 <h4>Package name</h4>
                 <p>{{ package.name }}</p>
+                {% if package.url %}
+                    <p>
+                        <small class="font-italic" style="font-size: 0.7em">
+                            <a href="{{ package.url }}" target="_blank" rel="nofollow noopener noreferrer">{{ package.url }}</a>
+                        </small>
+                    </p>
+                {% endif %}
```
akondas, line 29: What about something like this:
Screenshot from 2021-06-11 09-00-24

akondas:
```html
        <div class="row">
            <div class="col-3">
                <h4>Package name</h4>
                <p>{{ package.name }}</p>
            </div>
            <div class="col-3">
                <h4>Repository URL</h4>
                <p><a href="{{ package.url }}" target="_blank" rel="nofollow noopener noreferrer">{{ package.url }}</a></p>
            </div>
            <div class="col-6">
                <h4>Description</h4>
                <p>{{ package.description }}</p>
            </div>
        </div>
```

noniagriconomie: lets do this :)

src/Query/User/Model/PackageDetails.php
noniagriconomie: yes I have understood this
I was asking to, as fallback to avoid null, define name as url by default, and the sync may/will update it :)

akondas: we would prefer name to remain null if it is not present :wink:

noniagriconomie: I understood

akondas (approved):

akondas:
Thanks @noniagriconomie :medal_military:

PR #503, opened 2021-08-23
fix: Explicitly cast to boolean

eFrane:
With strict type checking, passing a non-boolean as email scan result leads to a fatal error.


akondas:
Hi, thanks for contribution. Could you tell us how `email_scan_result` can have other value than bool? This value comes from database and is always bool ... or maybe I don't known something?

eFrane:
It is not bool when using a sqlite database. I know that this is not the recommended configuration, but it can't hurt to cast it anyway in my opinion.

akondas (approved):

akondas:
Thanks :+1:

PR #531, opened 2021-12-31
Ugrade phpstan to ^1.3

alamirault:
Phpstan 1.0.0 was released in november.

This PR use last version ! 

We may also add `@deprecated` annotation on `getUsername` methods . It's deprecated in UserInterface since sf 5.3 in favor of getUserIdentifier()


alamirault:
With no use of deprectaed `getUsername` method coverage is now bellow 99%
`Code coverage is 98.87%, which is below the accepted 99%`

Should I add an unit test on `getUsername` ?

akondas:
> Should I add an unit test on getUsername ?

Yes please (in fact, this coverage is mainly for checking open source PR) ... or you can also lower it to 98% :shrug:

akondas (approved):

akondas (approved):

akondas:
Big thanks @alamirault :beers:

src/Security/Model/Organization.php
```diff
@@ -53,7 +53,7 @@ public function getUsername()
 
     public function getUserIdentifier(): string
     {
-        return $this->getUsername();
```
xvilo, line 56: Why did you make this change? I suppose we return the `Organization` alias here now, instead of the userIdentifier?

alamirault: `getUsername` is deprecated in symfony, phpstan extension for symfony now report it. In fact, `getUsername` return also alias so it's same

PR #535, opened 2022-01-05
Support redis symfony/messenger consumers

### LINKED ISSUE #533: Allow other symfony/messenger transports than doctrine
The repman environment file has a section for symfony/messenger to allow you to choose a transport by configuring a DSN.
https://github.com/repman-io/repman/blob/174c38190086e9ebf1f6d7849469742c8edd4e68/.env#L41-L46

However the symfony/messenger config yaml has the doctrine transport hardcoded instead of using this environment variable
https://github.com/repman-io/repman/blob/174c38190086e9ebf1f6d7849469742c8edd4e68/config/packages/messenger.yaml#L10-L12

Similarly only the `symfony/doctrine-messenger` is required in the composer.json file.
https://github.com/repman-io/repman/blob/174c38190086e9ebf1f6d7849469742c8edd4e68/composer.json#L58

It there any specific reason this was hardcoded to doctrine? I would like to use the [redis transport](https://symfony.com/doc/current/messenger.html#redis-transport) as it allows multiple workers consuming the same queue.

dheineman:
The pull requests adds support for using the symfony/messenger redis consumers. The default is still the doctrine but overriding the MESSENGER_TRANSPORT_* env variables allows users to switch to redis.

Initial testing indicates no problems but not finished yet.

Fixes https://github.com/repman-io/repman/issues/533

p.s. The added `allowed-plugins` in the composer.json is part of composer 2.2. Read more about it in the [release post](https://blog.packagist.com/composer-2-2/#more-secure-plugin-execution)

### DIFF
--- a/.env
+++ b/.env
@@ -41,8 +41,11 @@ DATABASE_URL=postgresql://127.0.0.1:5432/main?serverVersion=11&charset=utf8
 ###> symfony/messenger ###
 # Choose one of the transports below
 # MESSENGER_TRANSPORT_DSN=amqp://guest:guest@localhost:5672/%2f/messages
-# MESSENGER_TRANSPORT_DSN=doctrine://default
+# MESSENGER_TRANSPORT_FAILED_DSN=amqp://guest:guest@localhost:5672/%2f/failed
+MESSENGER_TRANSPORT_DSN=doctrine://default?auto_setup=false
+MESSENGER_TRANSPORT_FAILED_DSN=doctrine://default?queue_name=failed&auto_setup=false
 # MESSENGER_TRANSPORT_DSN=redis://localhost:6379/messages
+# MESSENGER_TRANSPORT_FAILED_DSN=redis://localhost:6379/failed
 ###< symfony/messenger ###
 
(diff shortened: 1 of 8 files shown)


h4ckroot:
Hello @dheineman I hope you are doing well. I know this is a bit old issue, but we are stuck with the same problem (can not scale consumers for repman, which causes sync delays). I just wanted to check if you managed to get this working? 
Thanks

dheineman:
> Hello @dheineman I hope you are doing well. I know this is a bit old issue, but we are stuck with the same problem (can not scale consumers for repman, which causes sync delays). I just wanted to check if you managed to get this working? Thanks

Hey @h4ckroot, it has indeed been a while since i created this pull request and i/we have since then moved away from repman due to it seemingly no longer being actively maintained and packagist starting to block requests due to inefficient mirroring (see https://github.com/repman-io/repman/issues/695) to https://github.com/vtsykun/packeton

From what i remember/see in the diff the change worked just fine but i don't know what else has changed in the last 4 years so i suggest testing yourself

PR #537, opened 2022-01-05
Performance: read and parse file contents after having send the response headers

Jeroeny:
When running `composer update` (or other commands that read the registry's package data), composer will fetch the `p2/<organization>/<package>.json` endpoint for every package. It will first read the response headers, to determine if the contents of the response should be downloaded. The contents of the response will only be downloaded if the local composer cache is outdated (the last-modified of the server is later then that of the client's files).  
If the cache is up-to-date, the rest of the response will not be downloaded (making the `update` command faster). 

However, server-side, repman always reads and parses the package data. Performance can possibly be improved by first determining the `lastModified` and ensuring the headers are output before the reading and parsing of the package data. 

Maybe I misinterpreted the code and this was already happening, but it didn't seem like it to me. Also this PR is not yet tested, but I'd like to get the proposal confirmed first. 

Docs reference:

https://getcomposer.org/doc/05-repositories.md

> Caching is done via the use of If-Modified-Since header, so make sure you return Last-Modified headers and that they are accurate.

### DIFF
--- a/src/Controller/RepoController.php
+++ b/src/Controller/RepoController.php
@@ -13,13 +13,13 @@ use Sensio\Bundle\FrameworkExtraBundle\Configuration\Cache;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Messenger\MessageBusInterface;
 use Symfony\Component\Routing\Annotation\Route;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
-use Symfony\Component\Routing\RouterInterface;
 
 final class RepoController extends AbstractController
 {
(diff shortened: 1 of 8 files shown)


src/Controller/RepoController.php
```diff
@@ -143,21 +144,22 @@ public function downloads(Request $request, Organization $organization): JsonRes
...
         );
 
-        if ($providerData === []) {
+        if ($lastModified === null) {
             throw new NotFoundHttpException();
         }
 
-        $response = (new JsonResponse($providerData))
+        $response = (new StreamedResponse(function () use ($loader): void {
+            echo \json_encode($loader(), JsonResponse::DEFAULT_ENCODING_OPTIONS);
```
xvilo, line 159: `echo` is a forbidden statement in phpstan, can we do this differently?

Jeroeny, line 159: Alternatively, this could use a stream output like `$outputStream = \fopen('php://output', 'wb');`.

Jeroeny: Alternatively, this could use a stream output like `$outputStream = \fopen('php://output', 'wb');`.

akondas:
But `lastModified` is set for each response, I thought it was enough :thinking: 

Can you provide some speed test (before and after) so we can see how it affects performance?

Generally, I am yes for this PR, but before we start polishing it, I would like to know how much it is worth

Jeroeny:
@akondas It was a bit of a challenge to setup locally, but I've got it now. I've done various tests with both parallel batches of requests and sequential. Though not always consistent, It seems the responsetime is up to ~5% less. E.g. when doing 100 requests, the average response time was 113ms vs 119ms. There could be more gain on setups with worse disk performance. Not sure how significant that is. But if a `composer update` checks 100+ packages, it may add up.

akondas:
in theory you are right, but I need a little more time to think about it, because the improvment is relatively small,
but good job anyway :wink:

src/Controller/RepoController.php
Jeroeny: Replaced `echo` with `fwrite()`

Jeroeny:
@akondas This might have more performance gains for setups with network-mounted disks. I think the change is worth to try. Could you reconsider the PR? I've updated the code to use `\fwrite()`, like the other `StreamResponse` implementations do too.

akondas:
Looks ok, we can give it a try. Do you have any benchmarks that can show how much we gain with this PR?

Jeroeny:
> Looks ok, we can give it a try. Do you have any benchmarks that can show how much we gain with this PR?

I currently do not have a setup to do extensive benchmarks, nor the time to set it up unfortunately.

PR #539, opened 2022-01-06
Make AWS auth credentials optional

### LINKED ISSUE #538: Bug: Opaque Auth environment variable is not optional
When setting storage to S3 and not supplying the Opaque Auth setting, it will give the following exception:

```
[WARNING] Some commands could not be registered:

In EnvVarProcessor.php line 172:

 [Symfony\Component\DependencyInjection\Exception\EnvNotFoundException]
 Environment variable not found: "STORAGE_AWS_OPAQUE_AUTH".
```

xvilo:
Fixes #538


config/services.yaml
```diff
@@ -19,7 +19,10 @@ parameters:
     security_advisories_db_repo: 'https://github.com/FriendsOfPHP/security-advisories.git'
     instance_id_file: '%kernel.project_dir%/var/instance-id'
     kernel_version: !php/const Buddy\Repman\Kernel::REPMAN_VERSION
-    storage.aws.endpoint: ~
+    storage.aws.endpoint: ~ # Default to null
+    storage.aws.key: ~
+    storage.aws.secret: ~
+    storage.aws.opaque_auth: false # Default to true
```
akondas, line 25: `false # Default to true` so true or false?

xvilo: Good catch. The default should be `false`, only when explicitly set to `true`, we also want the key and secret.

akondas (approved):

akondas:
Thanks :medal_sports:

PR #547, opened 2022-01-10
Feature: Ability for Instance admin to invite themselves as an Organization Admin

xvilo:
### What

This is sometimes a feature we would like to use to (temporarily) join a specific Organization within our self-hosted instance. It can be handy to do some quick debugging or checking. Now we have to ask the Organization admin to add us. 

### Flow

- On the Organizations admin overview page, press `Invite current user` (outdated image)
Screenshot 2022-01-10 at 13 38 00
- The user will then be invited as usual as an Organization `owner`
Screenshot 2022-01-10 at 13 37 47
- When the was already invited to the Organization, the button will not be visible
Screenshot 2022-01-10 at 13 47 24


xvilo:
@akondas can this be reviewed?

akondas:
Yes, I will definitely do it, I have more work planned for repman so it should be soon. Sorry for the delays, but I have a lot of other work to do.

PR #562, opened 2022-02-10
Fix detection of latest stable release (fix #443)

### LINKED ISSUE #443: RuntimeException - Version 9999999-dev for package [package-name] not found
I'm on the cloud version of Repman and one of our packages started reporting the error in the title for some reason.
Not sure where it tries to fetch that from, I can't find it anywhere in the repository at any rate.
I've tried re-syncing and updating the repo in Repman, nothing changes.

It also says that there is no stable version, even though there is one, it even says when it was released next to it, like this:

This is what that looks like

blmage:
If the first fetched package is not stable, but has a greater version than any of the stable packages, wrong behaviors such as that reported in #443 may occur.

### DIFF
--- a/src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
+++ b/src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
@@ -79,6 +79,13 @@ final class ComposerPackageSynchronizer implements PackageSynchronizer
             $latest = current($packages);
 
             foreach ($packages as $p) {
+                if ($p->getStability() === Version::STABILITY_STABLE) {
+                    $latest = $p;
+                    break;
+                }
+            }
+
+            foreach ($packages as $p) {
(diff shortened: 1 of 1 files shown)


src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
```diff
@@ -79,6 +79,13 @@ public function synchronize(Package $package): void
 
             $latest = current($packages);
 
+            foreach ($packages as $p) {
+                if ($p->getStability() === Version::STABILITY_STABLE) {
```
akondas, line 83: we can put this `if` inside existing loop, also please provided test for that case

giggsey:
Can I help with this to get it over the line?

akondas:
sorted in #617 feel free to reopen if I make false assumptions :wink:

PR #563, opened 2022-02-10
Truncate overly long `suggest` fields before persisting them

blmage:
The `constraint` field from the `organization_package_link` table is limited to 255 chars.

Currently, synchronization fails for packages that use `suggest` fields longer than 255 chars, and it becomes impossible to manage them from the UI (even if only to remove them).

Note that I am open to suggestions on how we might otherwise handle these cases.


alamirault:
Truncated description is not a problem ?

We can maybe change type in database in order to allow longest text ? And manage length in UI ? (css, ellipsis or something else)

akondas:
agree with @alamirault, you'd better change the field type to text :)

Issue #556, opened 2022-01-31
Abandon packages


giggsey:
It'll be useful to abandon packages, and highlight this on the UI (suggesting a replacement if one is supplied)

xvilo:
@giggsey has this been fixed with #567?

giggsey:
Yup, that'll do

PR #567, opened 2022-02-20
Show when package is deprecated

alamirault:
This PR allow to see deprecated packages on details. (#556)
deprecated

If it's ok for you I can also add it to the listing page, but I'm not sure for the placement and style

### DIFF
--- a/src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
+++ b/src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Buddy\Repman\Service\PackageSynchronizer;
 
 use Buddy\Repman\Entity\Organization\Package;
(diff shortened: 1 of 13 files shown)


src/Service/PackageSynchronizer/ComposerPackageSynchronizer.php
```diff
@@ -175,6 +176,13 @@ public function synchronize(Package $package): void
                         $package->addLink(new Link(Uuid::uuid4(), $package, 'suggests', $linkName, $linkDescription));
                         $encounteredLinks['suggests-'.$linkName] = true;
                     }
+
+                    // Is abandoned ?
+                    if ($latest instanceof CompletePackage && $latest->isAbandoned()) {
+                        $package->setReplacementPackage($latest->getReplacementPackage() ?? '');
```
akondas, line 182: this is tricky, so null means normal package, empty string means abandoned package without replacement, and non empty string means abandoned package with replacement

akondas (approved):

PR #575, opened 2022-03-17
Enable or disable security scans.

### LINKED ISSUE #555: Allow disabling security scan for certain projects
We have some old projects within Repman that are EOL/not used, which have security vulnerable dependencies.

It'll be useful to allow disabling security scans for projects (default to on).

shochdoerfer:
Fixes #555. A new field enableSecurityScan is introduced which allows to enable or disable the security scans of a package. It's turned on by default and can only be changed by editing the package information. I felt it might be good to not offer to set the flag when adding a new package. If you feel this should be added, let me know.

When the enableSecurityScan field is set to false, the package details view will look like this - the status "Scan disabled" is also displayed in the package overview:
details

The package edit form looks like this:
edit


.env.docker
```diff
@@ -27,7 +27,7 @@ APP_GITLAB_API_URL=https://gitlab.com
 ###< app ###
 
 ###> symfony/framework-bundle ###
-APP_ENV=prod
+APP_ENV=dev
```
karniv00l, line 30: 👀

shochdoerfer: Damn, my bad. Reverted the change.

xvilo:
We would like to see this on organizational level, is this something that can be added? At a quick glance this is not yet available with this PR

shochdoerfer:
@xvilo sure, can do that. But to be honest, I would love to get approval first or even get the PR merged before tackling this.

Only the question would be what the setting on the org level would control, I see 2 options:
 1. it just acts as a default value for the setting on the package level. When the setting gets changed on the org level, the setting of the org's packages won't be updated
 1. once the setting on the org level gets changed, it would change the setting for all packages of the org

I think I would prefer option 1 but I'd leave the choice up to you ;)

xvilo:
Would be great if you could add this (later) to a (new) PR! I guess it would be best to have it set as a default for all new packages and make them overridable in the package settings. However, I think it's a good feature to also allow bulk change of all packages in an org to enabled or disabled for example

shochdoerfer:
@xvilo sure, happy to improve the product ;) Looks like 2 features then, one to define the default setting for new packages and one to run the bulk update. Sounds doable.

shochdoerfer:
Is any support needed to get this merged?

akondas (approved):
LGTM :+1: 
@marmichalski will check and then we can merge this :wink:

src/MessageHandler/Security/ScanPackageHandler.php
```diff
@@ -28,6 +28,9 @@ public function __invoke(ScanPackage $message): void
         if (!$package instanceof Package) {
             return;
         }
+        if (!$package->isEnabledSecurityScan()) {
```
marmichalski, line 31: Could this be covered by a ScanPackageHandlerTest?

akondas:
Thanks :beers:

PR #579, opened 2022-04-06
Add Rector - Automatic upgrades and code refactoring

xvilo:
This PR aims to add Rector to this project. Rector instantly upgrades and refactors the PHP code of your application. It can help you in 2 major areas:  
- Instant Upgrades (yes, for PHP upgrade PR and Symfony update PR coming later)
- Automatic Refactoring (e.g. code quality issues and such)

This PR adds some default rules for just PHP Level with `LevelSetList::UP_TO_PHP_74`. It also automatically imports namespace classes and will remove unused imports too. 

The amount of rules/changes could be discussed, but this is the default set delivered by Rector. My intention is to create a PR that updates the project to PHP 8.0, this is includes the upgrades of packages (already available locally) and update to rector so the code will use the latest language features and automatically remove/fix deprecations and such.


xvilo:
@akondas Can you take some time to get this reviewed?

.php-cs-fixer.dist.php
```diff
@@ -31,11 +31,6 @@
         'trim_array_spaces' => true,
         'single_blank_line_before_namespace' => true,
         'yoda_style' => false,
-        'global_namespace_import' => [
```
marmichalski, line 34: why?

marmichalski: Could you please elaborate on this?

I don't think adding one tool should alter the behaviour of the other. This PR introduces changes to **175** where, at quick glance, most of them is the effect of changing cs-fixer rules.

marmichalski: If you want to keep auto import names, there seems to be a solution for "short classess", too:
```php
$parameters->set(Option::IMPORT_SHORT_CLASSES, false);
```

composer.json
```diff
@@ -184,6 +188,7 @@
         "tests": [
             "@check-cs",
             "@phpstan",
+            "@rector",
```
marmichalski, line 191: And update this here, if you choose to split the rector scripts:
```suggestion
            "@rector:check",
```

rector.php
```diff
@@ -0,0 +1,29 @@
...
+
+declare(strict_types=1);
+
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+use PhpCsFixer\Fixer\Import\NoUnusedImportsFixer;
+use Rector\Core\Configuration\Option;
+use Rector\Core\ValueObject\PhpVersion;
+use Rector\Set\ValueObject\LevelSetList;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
```
marmichalski, line 11: Version 0.12.21 has been released meanwhile, introducing `Rector\Config\RectorConfig`. Should you wish to update 🥲

rector.php
```diff
@@ -0,0 +1,29 @@
...
+    // here we can define, what sets of rules will be applied
+    // tip: use "SetList" class to autocomplete sets
+    $containerConfigurator->import(LevelSetList::UP_TO_PHP_74);
```
marmichalski, line 14: IMHO `SetList::CODE_QUALITY` is also useful.

rector.php
```diff
@@ -0,0 +1,29 @@
...
+
+    $parameters = $containerConfigurator->parameters();
+    $parameters->set(Option::AUTO_IMPORT_NAMES, true)
+        ->set(Option::PARALLEL, true)
+        ->set(Option::PHP_VERSION_FEATURES, PhpVersion::PHP_74)
+        ->set(Option::PATHS, [
+            __DIR__ . '/bin',
+            __DIR__ . '/src',
+            __DIR__ . '/tests'
+        ]);
```
marmichalski, line 28: Shouldn't there also be `Option::PHPSTAN_FOR_RECTOR_PATH` when phpstan is available in the project?

(You might need to create separate phpstan.neon config for rector, with just the extensions and level)

.php-cs-fixer.dist.php
akondas: Yes, agree with @marmichalski, pleas do not change cs-fixer rules here, this change is out of scope of this PR

rector.php
xvilo: It appears the default path is `phpstan.neon`, will not be adding it now. Have to investigate later what it does exactly :)

rector.php
xvilo: Done

rector.php
xvilo: Added, thats handy :)

xvilo:
Thanks, @marmichalski for your suggestions and comments, I've implemented these as best as I could. Can you do another review of this PR?

src/Controller/ProxyController.php
```diff
@@ -244,7 +244,7 @@ public function distribution(string $package, string $version, string $ref, stri
      */
     public function downloads(Request $request): JsonResponse
     {
-        $contents = \json_decode($request->getContent(), true);
+        $contents = \json_decode($request->getContent(), true, 512, JSON_THROW_ON_ERROR);
```
marmichalski, line 247: If it's rector, I think this rule should be skipped. If it's you, then you forgot to wrap it again after updating the PR. :stuck_out_tongue:

src/Controller/ProxyController.php
```diff
@@ -253,11 +253,7 @@ public function downloads(Request $request): JsonResponse
         }
 
         $this->messageBus->dispatch(new AddDownloads(
-            \array_map(function (array $data): Package {
-                return new Package($data['name'], $data['version']);
-            }, \array_filter($contents['downloads'], function (array $row): bool {
-                return isset($row['name'], $row['version']);
-            })),
+            \array_map(fn(array $data): Package => new Package($data['name'], $data['version']), \array_filter($contents['downloads'], fn(array $row): bool => isset($row['name'], $row['version']))),
```
marmichalski, line 256: It reads bad, could this be reformatted?
```suggestion
            \array_map(
                fn(array $data): Package => new Package($data['name'], $data['version']),
                \array_filter($contents['downloads'], fn(array $row): bool => isset($row['name'], $row['version']))
            ),
```

Both of the arrow functions could be marked as static (`static fn`), too, but that's just nitpicking.

src/Controller/RepoController.php
```diff
@@ -112,7 +112,7 @@ public function distribution(Organization $organization, string $package, string
      */
     public function downloads(Request $request, Organization $organization): JsonResponse
     {
-        $contents = json_decode($request->getContent(), true);
+        $contents = json_decode($request->getContent(), true, 512, JSON_THROW_ON_ERROR);
```
marmichalski, line 115: Missing try-catch (or not using throw flag).

src/Query/Admin/UserQuery/DbalUserQuery.php
```diff
@@ -73,7 +71,7 @@ private function hydrateUser(array $data): User
             $data['id'],
             $data['email'],
             $data['status'],
-            json_decode($data['roles'])
+            json_decode($data['roles'], null, 512, JSON_THROW_ON_ERROR)
```
marmichalski, line 74: Change in behaviour (different exception being thrown now if I'm correct)?

But I believe it's fine here.

src/Query/User/Model/ScanResult.php
```diff
@@ -61,6 +61,6 @@ public function isNotAvailable(): bool
      */
     public function content(): array
     {
-        return json_decode($this->content, true);
+        return json_decode($this->content, true, 512, JSON_THROW_ON_ERROR);
```
marmichalski, line 64: Again, I'd suggest to just not enforce throw flag.

I am aware that an exception thrown in that place, no matter if by decode returning false or throwing an error, should actually never happen and when it does the exception class is the slightest problem.

src/Security/ApiUserProvider.php
```diff
@@ -86,8 +87,8 @@ private function hydrateUser(array $data): User
             $data['status'],
             $data['email_confirmed_at'] !== null,
             $data['email_confirm_token'],
-            json_decode($data['roles'], true),
-            array_map(fn (array $data) => new User\Organization($data['alias'], $data['name'], $data['role'], $data['has_anonymous_access']), $organizations),
+            json_decode($data['roles'], true, 512, JSON_THROW_ON_ERROR),
```
marmichalski, line 90: :trumpet:

src/Security/UserProvider.php
```diff
@@ -101,8 +102,8 @@ private function hydrateUser(array $data): User
             $data['status'],
             $data['email_confirmed_at'] !== null,
             $data['email_confirm_token'],
-            json_decode($data['roles'], true),
-            array_map(fn (array $data) => new User\Organization($data['alias'], $data['name'], $data['role'], $data['has_anonymous_access']), $organizations),
+            json_decode($data['roles'], true, 512, JSON_THROW_ON_ERROR),
```
marmichalski, line 105: :drum:

src/Service/Organization/AliasGenerator.php
```diff
@@ -10,7 +10,7 @@ public function generate(string $name): string
     {
         $text = transliterator_transliterate('Any-Latin; Latin-ASCII; [^\u001F-\u007f] remove', $name);
         $text = strtolower((string) $text);
-        $text = preg_replace('~[^-\w]+~', ' ', $text);
+        $text = preg_replace('~[^\\-\w]+~', ' ', $text);
```
marmichalski, line 13: Was this added by rector? Both patterns work the same here, but this might be good to explicitly escape the `-` character, so :+1:.

src/Service/Proxy.php
```diff
@@ -281,7 +281,7 @@ private function decodeMetadata(string $package): array
     {
         /** @var Metadata $metadata */
         $metadata = $this->metadata($package)->getOrElse(Metadata::fromString('[]'));
-        $metadata = \json_decode((string) \stream_get_contents($metadata->stream()), true);
+        $metadata = \json_decode((string) \stream_get_contents($metadata->stream()), true, 512, JSON_THROW_ON_ERROR);
```
marmichalski, line 284: Change in behaviour. Either try-catch or skip the throw enforcing rule. There is an `is_array` check just below it.

tests/Functional/Controller/OrganizationControllerTest.php
```diff
@@ -480,8 +481,8 @@ public function testPackageDetails(): void
         $packageId = $this->fixtures->addPackage($buddyId, 'https://buddy.com');
         $versions = [
             new Version(Uuid::uuid4(), '1.0.0', 'someref', 1234, new \DateTimeImmutable(), Version::STABILITY_STABLE),
-            new Version(Uuid::uuid4(), '1.0.1', 'ref2', 1048576, new \DateTimeImmutable(), Version::STABILITY_STABLE),
-            new Version(Uuid::uuid4(), '1.1.0', 'lastref', 1073741824, new \DateTimeImmutable(), Version::STABILITY_STABLE),
+            new Version(Uuid::uuid4(), '1.0.1', 'ref2', 1_048_576, new \DateTimeImmutable(), Version::STABILITY_STABLE),
```
marmichalski, line 484: I would personally skip the `AddLiteralSeparatorToNumberRector` rule. Check with @akondas.

marmichalski (commented):
I guess it's good overall, some code style after rector could be fixed here and there, but nothing big.

I can only suggest to skip the rules:
* JsonThrowOnErrorRector
* AddLiteralSeparatorToNumberRector

PR #586, opened 2022-05-09
Add unit test for ScanPackageHandler

shochdoerfer:
As [discussed](https://github.com/repman-io/repman/pull/575), add unit test for the ScanPackageHandler.

### DIFF
--- a/tests/Unit/MessageHandler/Security/ScanPackageHandlerTest.php
+++ b/tests/Unit/MessageHandler/Security/ScanPackageHandlerTest.php
@@ -1,0 +1,89 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Buddy\Repman\Tests\Unit\MessageHandler\Security;
+
+use Buddy\Repman\Entity\Organization\Package;
+use Buddy\Repman\Message\Security\ScanPackage;
+use Buddy\Repman\MessageHandler\Security\ScanPackageHandler;
+use Buddy\Repman\Repository\PackageRepository;
+use Buddy\Repman\Service\Security\PackageScanner;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Ramsey\Uuid\Uuid;
+
+final class ScanPackageHandlerTest extends TestCase
+{
+    /**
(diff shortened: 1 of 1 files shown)


tests/Unit/MessageHandler/Security/ScanPackageHandlerTest.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    private PackageRepository $packageRepository;
+
+    protected function setUp(): void
+    {
+        $this->packageScanner = $this->createMock(PackageScanner::class);
+        $this->packageRepository = $this->createMock(PackageRepository::class);
+    }
+
+    public function testHandlerIgnoresNonExistentPackages(): void
```
marmichalski, line 33: This is covered here: https://github.com/repman-io/repman/blob/8b1d42ea876ee011b236791238a5d30420c83b2c/tests/Integration/MessageHandler/Security/ScanPackageHandlerTest.php#L40

marmichalski (commented):
I believe it is fine how it is, but is there any specific reason you have decided to unit test this instead of adding more scenarios to already existing [integration test](https://github.com/repman-io/repman/blob/8b1d42ea876ee011b236791238a5d30420c83b2c/tests/Integration/MessageHandler/Security/ScanPackageHandlerTest.php)?

tests/Unit/MessageHandler/Security/ScanPackageHandlerTest.php
shochdoerfer: The simple answer is: I did not expect this to be covered by an integration test ;) I checked the unit tests, found none, and added the test case. Surely, I can migrate them over the integration tests.

shochdoerfer: I checked the unit test and the integration test. I don't think it makes sense to merge them. The integration test currently "just" checks for errors (exceptions), the unit test is testing the flow in the handler. I don't think I could test the flow in an integration test (e.g. the code currently has 2 return statements. I don't think I could distinguish between both in the integration test).

akondas (approved):
We can merge this, thanks

PR #591, opened 2022-05-21
Introduce org setting enable_security_scan

shochdoerfer:
As discussed in #575 introduce a setting for the organization to define the default for new packages.

When a new organization is created the default value for the `enable_security_scan` configuration is true to stick to the previous defaults. The setting can be changed afterward in the respective form like this:
scan

When a new package is created the default value for `enable_security_scan` flag on the package level is set to the value defined in the organization settings.

As you can see in the screenshot above, changing the setting on the organization level does not change the settings for all packages of that organization. It just will change the default value for newly created packages.

### DIFF
--- a/src/Entity/Organization.php
+++ b/src/Entity/Organization.php
@@ -70,6 +70,11 @@ class Organization
      */
     private bool $hasAnonymousAccess = false;
 
+    /**
+     * @ORM\Column(type="boolean", options={"default": true})
+     */
+    private bool $enable_security_scan = true;
+
     public function __construct(UuidInterface $id, User $owner, string $name, string $alias)
     {
         $this->id = $id;
@@ -254,6 +259,11 @@ class Organization
     public function changeAnonymousAccess(bool $hasAnonymousAccess): void
     {
         $this->hasAnonymousAccess = $hasAnonymousAccess;
+    }
+
+    public function enableSecurityScan(bool $enableSecurityScan): void
+    {
+        $this->enable_security_scan = $enableSecurityScan;
     }
 
     private function isLastOwner(User $user): bool
--- a/src/Form/Type/Organization/EnableSecurityScanType.php
+++ b/src/Form/Type/Organization/EnableSecurityScanType.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Buddy\Repman\Form\Type\Organization;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+class EnableSecurityScanType extends AbstractType
+{
+    public function getBlockPrefix(): string
+    {
+        return '';
+    }
+
+    /**
+     * @param array<mixed> $options
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('isSecurityScanEnabled', ChoiceType::class, [
+                'choices' => [
+                    'Enable' => true,
+                    'Disable' => false,
+                ],
+                'label' => 'Enable security scan for new packages',
+                'required' => true,
+            ])
+            ->add('enableSecurityScan', SubmitType::class, ['label' => 'Change'])
+        ;
+    }
+}
--- a/src/Message/Organization/EnableSecurityScan.php
+++ b/src/Message/Organization/EnableSecurityScan.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Buddy\Repman\Message\Organization;
+
+final class EnableSecurityScan
+{
+    private string $organizationId;
+    private bool $enableSecurityScan;
+
+    public function __construct(string $organizationId, bool $enableSecurityScan)
+    {
+        $this->organizationId = $organizationId;
+        $this->enableSecurityScan = $enableSecurityScan;
+    }
+
+    public function organizationId(): string
+    {
+        return $this->organizationId;
+    }
+
+    public function hasSecurityScanEnabled(): bool
+    {
+        return $this->enableSecurityScan;
+    }
+}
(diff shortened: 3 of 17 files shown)


src/Message/Organization/EnableSecurityScan.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Buddy\Repman\Message\Organization;
+
+final class EnableSecurityScan
```
akondas, line 7: hmm, this command can enable and disable security scan, so maybe we can pick better name, for example: `ChangeSecurityScan` or `UpdateSecurityScanOption` ... wdyt?

akondas (commented):
Looks good

src/Message/Organization/EnableSecurityScan.php
shochdoerfer: Yeah, I was a bit unsure how to properly name it. I felt it would be good to name the message like the flag I introduced, but since the message will toggle the flag it sounds a bit wired.

We can surely go with `ChangeSecurityScan` - I'll think about it, maybe I can come up with a better name. Will push later today...

shochdoerfer: Went with `ChangeSecurityScanConfiguration`.

src/Entity/Organization.php
```diff
@@ -70,6 +70,11 @@ class Organization
      */
     private bool $hasAnonymousAccess = false;
 
+    /**
+     * @ORM\Column(type="boolean", options={"default": true})
+     */
+    private bool $enable_security_scan = true;
```
marmichalski, line 76:
```suggestion
    private bool $enableSecurityScan = true;
```

or maybe:
```suggestion
    private bool $hasEnabledSecurityScan = true;
```
?

src/Entity/Organization.php
```diff
@@ -256,6 +261,11 @@ public function changeAnonymousAccess(bool $hasAnonymousAccess): void
         $this->hasAnonymousAccess = $hasAnonymousAccess;
     }
 
+    public function enableSecurityScan(bool $enableSecurityScan): void
```
marmichalski, line 264:
```suggestion
    public function changeSecurityScanAvailability(bool $enabled): void
```
?

src/Form/Type/Organization/EnableSecurityScanType.php
```diff
@@ -0,0 +1,36 @@
...
+declare(strict_types=1);
+
+namespace Buddy\Repman\Form\Type\Organization;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+class EnableSecurityScanType extends AbstractType
```
marmichalski, line 12:
```suggestion
final class EnableSecurityScanType extends AbstractType
```

src/Query/User/Model/Organization.php
```diff
@@ -93,4 +95,9 @@ public function hasAnonymousAccess(): bool
     {
         return $this->hasAnonymousAccess;
     }
+
+    public function isSecurityScanEnabled(): bool
```
marmichalski, line 99: Could all methods/properties referring security scan be unified?

templates/organization/settings.html.twig
```diff
@@ -47,6 +47,21 @@
         {{ form_end(anonymousAccessForm) }}
     <hr />
 
+    <h4>Package Security scan</h4>
+    <p>
+        Enabling or disable the security scan for new packages. Changing this setting will not affect existing packages!
```
marmichalski, line 52:
```suggestion
        Enable or disable the security scan for new packages. Changing this setting will not affect existing packages!
```

src/MessageHandler/Organization/AddPackageHandler.php
```diff
@@ -29,7 +29,8 @@ public function __invoke(AddPackage $message): void
                     $message->type(),
                     $message->url(),
                     $message->metadata(),
-                    $message->keepLastReleases()
+                    $message->keepLastReleases(),
+                    $message->hasSecurityScanEnabled()
```
marmichalski, line 33: You have the organization here in the handler, you do not have to pass this feature through a message object.
```php
$organization = $this->organization->getById(Uuid::fromString($message->organizationId()));
$organization->addPackage(new Package(..., $organization->hasEnabledSecurityScan()));
```

src/MessageHandler/Organization/EnableSecurityScanHandler.php
```diff
@@ -0,0 +1,28 @@
...
+declare(strict_types=1);
+
+namespace Buddy\Repman\MessageHandler\Organization;
+
+use Buddy\Repman\Message\Organization\ChangeSecurityScanConfiguration;
+use Buddy\Repman\Repository\OrganizationRepository;
+use Ramsey\Uuid\Uuid;
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+
+final class EnableSecurityScanHandler implements MessageHandlerInterface
```
marmichalski, line 12:
```suggestion
final class ChangeSecurityScanConfigurationHandler implements MessageHandlerInterface
```

src/Entity/Organization.php
shochdoerfer: Renamed to $enableSecurityScan

src/Form/Type/Organization/EnableSecurityScanType.php
shochdoerfer: Made class final.

templates/organization/settings.html.twig
shochdoerfer: Fixed.

src/Query/User/Model/Organization.php
shochdoerfer: I can't follow. What do you mean by "unified"? Can you give a better example?

marmichalski: Mix of `is` and `has`

Jarzebowsky:
Any update on this?

shochdoerfer:
good question ;) Sadly I completely lost track of this issue due to a lot of other things I had to focus on.

Let me know which of the findings need to be fixed before the PR can be merged and I'll see that I can take care soonish.

PR #609, opened 2022-09-06
Fix data pull for security advisories database

xvilo:


giggsey:
https://github.com/FriendsOfPHP/security-advisories/branches doesn't have a main branch?

akondas:
Yes, can you explain?
Screenshot from 2022-09-10 14-05-52

PR #612, opened 2022-09-08
Configure PHPStan paths

xvilo:
This configures the paths to check too in `phpstan.neon`. Makes it easier to just run the command, and for IDEs to auto check your code while developing

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -164,7 +164,7 @@
         ],
         "phpstan": [
             "bin/console cache:clear --env=test",
-            "phpstan analyse src tests --level=max"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -164,7 +164,7 @@
         ],
         "phpstan": [
             "bin/console cache:clear --env=test",
-            "phpstan analyse src tests --level=max"
+            "phpstan analyse --level=max"
```
akondas, line 167: I think `level` can also be in config, right?

akondas (approved):

PR #617, opened 2022-09-20
Fix no stable release sorting bug

marickvantuil:
This fixes the issue in #562 where if the first fetched package is not stable, then no stable release can be determined because the version comparison will fail with `Comparator::greaterThan`.


aleho:
This actually fixes the issue on my installation.

akondas (approved):

PR #637, opened 2023-03-07
reduce docker image size by ~200MB

mikk150:
Right now repman docker container contains dependencies to build php modules, that really are not needed on runtime
Cleanup composer cache as well, as this too is not needed on runtime


akondas (approved):
thanks :beers:

PR #510, opened 2021-08-26
PHP 8.0 support

akondas:

PR #639, opened 2023-03-13
Use latest php version in buddy

slappyslap:
Fix build pipeline


marmichalski:
Updated as part of #510
(that pipeline currently is unavailable anyway)

PR #640, opened 2023-03-13
Fix: invalid package doesn't show if search is empty

slappyslap:
Hello,
This fix the package search and allow invalid package to show if search is empty


floverdevel:
i would love to see this bug fixed :)

slappyslap:
Me too, but the repository appears to be abandoned

PR #646, opened 2023-04-26
fix: Packagist proxy through HTTP Proxy

ymorocutti:
Use StreamContextFactory from composer package to create stream context as this method already support HTTP proxy settings


akondas:
Can we add test for that or it will be too complicated?

ymorocutti:
> Can we add test for that or it will be too complicated?

I could unit test method createContext with fake URL and multiple proxy/no_proxy settings to make sure options returned to create a context are what we expect.

akondas:
> > Can we add test for that or it will be too complicated?
> 
> I could unit test method createContext with fake URL and multiple proxy/no_proxy settings to make sure options returned to create a context are what we expect.

Works for me, at least test will concrete solution :sweat_smile:

PR #679, opened 2024-05-29
Fix #678 Allow configuration trustedHeaderSet

### LINKED ISSUE #678: AWS ELB offloading, impossible link with Oauth Provider
Request for behind AWS Elastic Load Balancer with SSL offloading are dropping the `s` in `https` protocol.

So during installation / setup of the first user. The url generated for the Oauth provider (bitbucket in our case) isn't a valid redirect_uri.

The `TRUSTED_PROXIES` environment variable is a good starting point. But without the TrustedHeader set to `Request::HEADER_X_FORWARDED_AWS_ELB`, it is impossible to get a proper uri.

pascallap:
Fix for #678 
Allow a configurable trustedHeaderSet.


pascallap:
No need after all.
Only the proper trusted was enought.... sorry for the trouble

PR #715, opened 2026-09-12
fix: prefer VCS tag over composer.json version when syncing

s00d:
## Summary
- If `composer.json` contains a hardcoded `"version"` that does not match the git tag, Composer **silently skips** that tag.
- Sync and webhook still look fine (202 / no error), but the release never shows up in Repman — you publish a release and nothing happens.
- For VCS packages, ignore `composer.json` `version` and take the version from the tag/branch name (as Composer docs recommend for VCS).

## Test plan
- [ ] Unit: `IgnoreComposerJsonVersionVcsDriverTest`
- [ ] Unit: `ComposerPackageSynchronizerTest::testSynchronizeGitPackagePrefersTagOverComposerJsonVersion` (repo with `"version": "1.0.0"` + tags `1.0.0` / `1.0.1`)
- [ ] Existing path/artifact sync tests still pass

Made with [Cursor](https://cursor.com)


s00d:
Fixed Buddy/phpstan failures: typed VcsDriver wrapper methods, removed `exec` from tests.

